PR #2, opened 2019-09-09
Project setup

pshemass:
C&P from zio-keeper. @mijicd please verify this.


pshemass:
@mijicd is on his way.

Issue #1, opened 2019-09-09
Adding logging environmental effect


jdegoes:
Here's a quick sketch of a logging API.

```scala
trait Logging {
  def logging: Logging.Service[Any]
}
object Logging {
  trait Service[-R] {
    // TODO: Define methods here
    def log(line: => String): ZIO[R, Nothing, Unit]
    def info(line: => String): ZIO[R, Nothing, Unit]
    def warning(line: => String): ZIO[R, Nothing, Unit]
    def error(line: => String): ZIO[R, Nothing, Unit]
    def errorCause(cause: Cause[Any]): ZIO[R, Nothing, Unit]
  }
}

package object logging extends Logging.Service[Logging] {
  ...
}
```

I don't think this is quite sufficient. We probably need to think about `FiberRef` integration and making sure we have a small set of operations that can be implemented by many different logging libraries (e.g. LogStage, all the ZIO loggers, slf4j, etc.).

softinio:
This is a good idea. Are we trying to get this done in time for release 1.0?

pshirshov:
This wouldn't work for our structured logging solution. I would propose this as a starting point:

```scala
trait Logging[F[_], Message] {
 def info(message: Message): F[Unit] 
  //... 
}

trait UnstructuredLogging[F[_]] extends Logging[F, String] 


// `Message` is a data structure with an implicit materializer from LogStage

trait StructuredLogging[F[_]] extends Logging[F, Message] {

}
```

NeQuissimus:
Looks remarkably similar to https://github.com/NeQuissimus/zio-slf4j/blob/master/src/main/scala/logger.scala ?

I use https://github.com/NeQuissimus/zio-slf4j/blob/master/src/main/scala/package.scala#L8 to turn any `A` into something log-able...

pshemass:
@NeQuissimus yes it is very similar. we would like to have very simple interface that can be back up with slf4j and other implementations. Moreover we would like to avoid dependencies to scalaz if possible.

NeQuissimus:
Oh, I know... I meant to point out that you are welcome to take whatever I have in that repo.
It would come down to implementations of `Logger.Service` on how each framework handles the actual logging, of course.

pshemass:
@pshirshov why we need ```F[_]``` ? I was thinking that ```ZIO``` would be enough in this case. We try to keep this interface very simple.

pshirshov:
[...]

PR #6, opened 2019-09-12
#1 Basic interface

pshemass:


core/src/main/scala/zio/logging/Logging.scala
```diff
@@ -0,0 +1,39 @@
...
+    def context: LoggingContext[Message]
+    def trace(message: Message): ZIO[R, Nothing, Unit]
+    def debug(message: Message): ZIO[R, Nothing, Unit]
+    def info(message: Message): ZIO[R, Nothing, Unit]
+    def warning(message: Message): ZIO[R, Nothing, Unit]
+    def error(message: Message): ZIO[R, Nothing, Unit]
+    def error(message: Message, cause: Cause[Any]): ZIO[R, Nothing, Unit]
+  }
+
+  final class LoggingContext[Message] private (ctx: FiberRef[Map[String, Message]]) {
```
thiloplanz: I think the context entries are not typed `Message`.  In Logstage, the context map is a `Map[String, Any]`, whereas `Message` is a structured log entry case class.

Maybe we need another generic type `Context` (which `trait Logging` can probably also fix to `String`)?

core/src/main/scala/zio/logging/Logging.scala
```diff
@@ -0,0 +1,39 @@
...
+
+trait AbstractLogging[Message] {
+  def logging: AbstractLogging.Service[Any, Message]
+}
+
+object AbstractLogging {
+
+  trait Service[-R, Message] {
+    def context: LoggingContext[Message]
+    def trace(message: Message): ZIO[R, Nothing, Unit]
```
thiloplanz: Should `message` be lazy? So that it does not need to be generated when the logger is not enabled for `trace`?

core/src/main/scala/zio/logging/Logging.scala
pshemass: my thinking is that those items from context you should be able to put in log statement as placeholders so it need to be align with ```Message```

core/src/main/scala/zio/logging/Logging.scala
pshemass: good point

core/src/main/scala/zio/logging/Logging.scala
jdegoes: Why have a separate type for `LoggingContext`? i.e. why not just inline the `FiberRef` into the `Service`?

core/src/main/scala/zio/logging/Logging.scala
```diff
@@ -0,0 +1,39 @@
...
+import zio.{ Cause, FiberRef, UIO, ZIO }
+
+trait AbstractLogging[Message] {
+  def logging: AbstractLogging.Service[Any, Message]
+}
+
+object AbstractLogging {
+
+  trait Service[-R, Message] {
+    def context: LoggingContext[Message]
```
jdegoes: I might call this `loggingContext` just to disambiguate.

core/src/main/scala/zio/logging/Logging.scala
jdegoes: If we need context to support values of any type, then we should not use `Any`. Instead, we should use something similar to ZIO Test `TestAnnotation`, which allows typed entries.

The question is:

 - Should we fix the context values to some type, like String or Message, or a new type parameter `Context`?
 - Or should we allow them to vary by key, like `TestAnnotationMap` / `TestAnnotation` in ZIO Test, in which case, we need a slightly different design here?

pshemass: > Why have a separate type for LoggingContext? i.e. why not just inline the FiberRef into the Service?

I believe that dealing with FiberRef it maybe not flexible enough and in cases that you need to even add something maybe a little bit cumbersome.

> Should we fix the context values to some type, like String or Message, or a new type parameter Context? 

what is a use case to add something to context in different type than ```Message```? those variables will end up in log file so they need to be in the same type as ```Message```  otherwise someone will need to provide those ugly converters. 

> Or should we allow them to vary by key, like TestAnnotationMap / TestAnnotation in ZIO Test, in which case, we need a slightly different design here?

I like that idea :)

core/src/main/scala/zio/logging/Logging.scala
pshemass: good point

core/src/main/scala/zio/logging/Logging.scala
jdegoes: Here is a design:

```scala
trait ContextMap {
  def get[V](key: ContextKey[V]): FiberRef[V]
}
class ContextKey[V] private (
  val identifier: String,
  val initial: V,
  val combine: (V, V) => V,
  private val classTag: ClassTag[V]
) { ... }
```

Then you can have:

```scala
def addContext[V](contextKey: ContextKey[V], v: V): ZIO[R, Nothing, Unit]
def getContext[V](contextKey: ContextKey[V]): ZIO[R, Nothing, Unit]
```

Then perhaps last thing is `span` like operator which can aggregate context:

```scala
def span[R, E, A](name: Message)(zio: ZIO[R, E, A]): ZIO[R, E, A]
```

So you could do:

```scala
span("database access") {
   ... all database access ...
}
```

I'm not 100% sure about `span` but the contextual methods seem nice / useful.

pshemass: I will create ticket for ```span``` because that might be good idea. 

I replaced ```LoggingContext``` with ```ContextMap``` let me know what do you think about that.

core/src/main/scala/zio/logging/Logging.scala

mijicd:
This PR is under review for some time now :). 

I propose to bump ZIO version to the latest one and merge if there are no major issues. Then spawn follow-ups to proceed. What do you think guys?

@pshemass @jdegoes

jdegoes:
I'll get on this ASAP

PR #10, opened 2019-11-24
add AbstractLogging#span (wip)

mschuwalow:
Allows allows contextMap to be merged on fiberjoin.
For every key that was set in the joinee a merge will done according to the contextkey

/cc @mijicd


pshemass:
Could you post here quick example so I will add this to README?

mschuwalow:
> Could you post here quick example so I will add this to README?

Sure. Something like:
```scala
import zio._
import zio.console._

object example {

  val allocated = ContextKey[Set[String]]("resouces", Set(), _ ++ _)

  val resource: ZManaged[Any, Nothing, Unit] = ZManaged.succeed(())

  val test =
    ZIO.environment[Logging].map(_.logging).flatMap { logger =>
      resource.use_ {
        logger.span(allocated, Set("outerResource")) {
          resource.use_ {
            logger.span(allocated, Set("innerResource")) {
              logger.loggingContext.get(allocated).flatMap(a => putStrLn(a.toString)) // Set("outerResource", "innerResource")
            }
          }
          logger.loggingContext.get(allocated).flatMap(a => putStrLn(a.toString)) // Set("outerResource")
        }
      }
    }
}

```

mschuwalow:
Thinking about this some more, I think the only way to properly support span (and remove and set by extension) is to have the ContextMap essentially be eventsourced. 

My idea is that per ContextKey we essentially have 3 operations {Add (add a new value -> merge with key.combine), Set (overwrite existing value), Remove (remove value)} that can be performed on it. During fiberjoin we find the changes per key that the joiner has not yet seen and merge them into the map.

For span we can apply a local change that is not reflected in the changelog and at the end of the function merge the changelog into the original map. That way we preserve all local changes, but can introduce 'temporary' value for span

PR #13, opened 2019-11-26
slf4j implementation

pshemass:
slf4j implementation with MDC

### DIFF
--- a/core/src/main/scala/zio/logging/Logging.scala
+++ b/core/src/main/scala/zio/logging/Logging.scala
@@ -1,8 +1,6 @@
 package zio.logging
 
-import zio.{ Cause, FiberRef, UIO, ZIO }
-
-import scala.reflect.ClassTag
+import zio.{Cause, ZIO}
 
 trait AbstractLogging[Message] {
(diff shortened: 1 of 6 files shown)


core/src/main/scala/zio/logging/Logging.scala
```diff
@@ -11,52 +9,14 @@ trait AbstractLogging[Message] {
 object AbstractLogging {
 
   trait Service[-R, Message] {
-    def loggingContext: ContextMap
+    def addToContext(key: String, value: Message): ZIO[R, Nothing, Unit]
```
jdegoes: This is not a very positive change in my opinion. Logging should not acquire characteristics of SLF4J.

pshemass: I'm trying to make this as simple as possible. if we will go with previous approach it gives a little bit too much flexibility and adding complexity. Moreover people will start using this context for more that logging and I don't like that. 

this will work perfectly for structure logging too. so it's not specific for slf4j. 

@jdegoes let me know if you have any particular example that this will limit us

PR #14, opened 2019-11-26
slf4j implemetation

pshemass:
this is using contextMap implemented before and similar concept as log4zio without contramap but we can easily added.

```
object Examples extends zio.App {
  val correlationId = ContextKey[Option[String]]("correlationId", None)

  val env =
    for {
      ctxMap <- ContextMap.empty
    } yield {
      val stringFormat = "[correlation-id = %s] %s"
      new Logging with LoggingContext with LoggingFormat[String] {
        self =>
        override def logging: AbstractLogging.Service[Any] = new Slf4jLogger()

        override def loggingContext: LoggingContext.Service[Any] = ctxMap

        override def format(message: String): ZIO[Any, Nothing, String] =
          loggerContext
            .get(correlationId)
            .map(correlationId => stringFormat.format(correlationId.getOrElse("undefined"), message))
            .provide(self)
      }
    }

    (for {
      fiber <- loggerContext.add(correlationId, Some("1234")).fork
      _ <- logger.info("info message without correlation id")
      _ <- fiber.join
      _ <- logger.info("info message with correlation id")
      _ <- logger.info("another info message with correlation id").fork
    } yield 1).provideSomeM(env)
}
```


core/src/main/scala/zio/logging/ContextMap.scala
```diff
@@ -2,7 +2,21 @@ package zio.logging
 
 import zio._
 
-final class ContextMap private (private val map: FiberRef[ContextMap.CMap]) {
+trait LoggingContext {
```
jdegoes: This was pulled out merely to support SLF4j. It's making the whole interface more complicated only to support one particular implementation.

IMO I don't think that's a good design move. Either logging Context is a good idea in which case we should emulate it for SLF4J (which is relatively straightforward), or it's not a good idea, in which case we should remove it.

core/src/main/scala/zio/logging/ContextMap.scala
```diff
@@ -2,7 +2,21 @@ package zio.logging
...
+  def loggingContext: LoggingContext.Service[Any]
+}
+
+object LoggingContext {
+  trait Service[-R] {
+    def get[V](key: ContextKey[V]): ZIO[R, Nothing, V]
+    def add[V](key: ContextKey[V], value: V): ZIO[R, Nothing, Unit]
+    def set[V](key: ContextKey[V], value: V): ZIO[R, Nothing, Unit]
```
jdegoes: The difference between add and set is confusing. I think it should only be `add` and the particular `ContextKey` can decide whether to "append" or "replace".

core/src/main/scala/zio/logging/ContextMap.scala
```diff
@@ -2,7 +2,21 @@ package zio.logging
...
+    def remove(key: ContextKey[_]): ZIO[R, Nothing, Unit]
+    def span[R1 <: R, E, A, V](key: ContextKey[V], value: V)(zio: ZIO[R1, E, A]): ZIO[R1, E, A]
+  }
+}
+
+final class ContextMap private (private val map: FiberRef[ContextMap.CMap]) extends LoggingContext.Service[Any] {
```
jdegoes: This needs to be in its own file.

core/src/main/scala/zio/logging/LoggingFormat.scala
```diff
@@ -0,0 +1,6 @@
+package zio.logging
+
+import zio._
+trait LoggingFormat[Message] {
```
jdegoes: Why does this exist? It is a proliferation of interfaces.

core/src/main/scala/zio/logging/AbstractLogging.scala
```diff
@@ -2,24 +2,20 @@ package zio.logging
 
 import zio.{ Cause, ZIO }
 
-trait AbstractLogging[Message] {
-  def logging: AbstractLogging.Service[Any, Message]
+trait AbstractLogging {
```
jdegoes: Why were all the type parameters deleted? LogStage specifically requires a non-String message type.

core/src/main/scala/zio/logging/ContextMap.scala
pshemass: This was not pulled to support SLF4j.... this was propose by your initial design. The only different is that you propose to tight interface to ```FiberRef[Map[String, Any]]``` and this is just a wrapper for that with context key that you propose to implement based on zio tests.

core/src/main/scala/zio/logging/ContextMap.scala
pshemass: I will remove ```add```

core/src/main/scala/zio/logging/LoggingFormat.scala
pshemass: this is the idea from ```log4zio``` that you mentioned in the issue that might be interesting. in ```log4zio``` that would be ```LogMedium``` in the future we can add those ```contramap``` to convert one to another.

pshemass: let me know I can get rid of this and back to simple interface

core/src/main/scala/zio/logging/AbstractLogging.scala
pshemass: instead of param here you can change ```LogFormat``` and one part of application can use different ```LogFormat```s

core/src/main/scala/zio/logging/ContextMap.scala
pshemass: sure

core/src/main/scala/zio/logging/LoggingFormat.scala
pshemass: This format also remove coupling to MDC because it let us control format of message.

```
override def format(message: String): ZIO[Any, Nothing, String] =
          loggerContext
            .get(correlationId)
            .map(correlationId => stringFormat.format(correlationId.getOrElse("undefined"), message))
            .provide(self)
```

PR #31, opened 2020-01-18
Feature/new design

pshemass:
example looks like that now 

```
object Examples extends zio.App {
  val correlationId = LogAnnotation[String](
    name = "correlationId",
    neutral = "undefined-correlation-id",
    combine = (_, newValue) => newValue,
    render = identity
  )

  val logFormat = "[correlation-id = %s] %s"

  val env =
    Slf4jLogger.make(
      LogLevel.Debug,
      (context, message) => logFormat.format(context.get(correlationId), message)
    )

    (for {
      fiber <- locallyAnnotate(correlationId, "1234")(ZIO.unit).fork
      _     <- log("info message without correlation id")
      _     <- fiber.join
      _ <- locallyAnnotate(correlationId, "1234111") {
            log("info message with correlation id") *>
              log(LogLevel.Error)("another info message with correlation id").fork
          }
    } yield 1).provideSomeM(env)
}
```


core/src/main/scala/zio/logging/Logger.scala
```diff
@@ -0,0 +1,64 @@
...
+      case NoLocation(_)                  => None
+    }
+
+  /**
+   * Constructs a logger provided the specified sink.
+   */
+  def make(logLevel: LogLevel, logger: (LogContext, => String) => UIO[Unit]): UIO[Logger] =
+    FiberRef
+      .make(LogContext.empty)
+      .flatMap(ref => ref.update(ctx => ctx.annotate(LogAnnotation.Level, logLevel)).as(ref))
```
jdegoes: Simplify to `.make(LogContext.empty.annotate(...))`

core/src/main/scala/zio/logging/Logger.scala
```diff
@@ -0,0 +1,64 @@
...
+  private[logging] def classNameForLambda(lambda: => AnyRef) =
+    tracing.tracer.traceLocation(() => lambda) match {
+      case SourceLocation(_, clazz, _, _) => Some(clazz)
+      case NoLocation(_)                  => None
+    }
+
+  /**
+   * Constructs a logger provided the specified sink.
+   */
+  def make(logLevel: LogLevel, logger: (LogContext, => String) => UIO[Unit]): UIO[Logger] =
```
jdegoes: This is like a "default log level". I'm not sure if that's good or not. But anyway, you can generalize it by changing it to `LogContext`, and then the user can supply defaults for everything, if they really want. Maybe default it to `LogContext.empty`

core/src/main/scala/zio/logging/Logger.scala
```diff
@@ -0,0 +1,64 @@
...
+          def locally[R, E, A](f: LogContext => LogContext)(zio: ZIO[R, E, A]): ZIO[R, E, A] =
+            ref.get.flatMap(context => ref.locally(f(context))(zio))
+
+          def log(line: => String): UIO[Unit] = ref.get.flatMap(context => logger(context, line))
+
+          def logContext: UIO[LogContext] = ref.get
+        }
+      }
+
+  val noopLogger =
```
jdegoes: Rename to `noop` or `nothing` or maybe `ignore`.

core/src/main/scala/zio/logging/Logger.scala
```diff
@@ -0,0 +1,64 @@
...
+    make(LogLevel.Off, (_, _) => ZIO.unit)
+
+  def consoleLogger(logLevel: LogLevel, format: (LogContext, => String) => String) =
```
jdegoes: Rename to `console`.

core/src/main/scala/zio/logging/Logger.scala
jdegoes: Generalize `make` to `def make[R](logLevel: LogLevel, logger: (LogContext, => String) => URIO[R, Any]): URIO[R, Logger]`. 

This will clean up some code down below.

core/src/main/scala/zio/logging/Logger.scala
```diff
@@ -0,0 +1,64 @@
...
+          (for {
+            date  <- currentDateTime
+            level = context.get(LogAnnotation.Level)
+            loggerName = context.get(LogAnnotation.Name) match {
+              case Nil => classNameForLambda(line).getOrElse("ZIO.defaultLogger")
+              case _   => LogAnnotation.Name.render(context.get(LogAnnotation.Name))
+            }
+            _ <- putStrLn(date.toString + " " + level.render + " " + loggerName + " " + format(context, line))
+
+          } yield ()).provide(env)
```
jdegoes: This can be deleted when you complete the above generalization.

core/src/main/scala/zio/logging/Logger.scala
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * A logger of strings.
+ */
+trait Logger extends LoggerLike[String]
+object Logger {
+
+  private val tracing = Tracing(Tracer.globallyCached(new AkkaLineNumbersTracer), TracingConfig.enabled)
+
+  private[logging] def classNameForLambda(lambda: => AnyRef) =
```
jdegoes: This will be a very slow way to get the enclosing class of the lambda. I'm not saying _don't_ do it (maybe it is useful enough, I am not sure), but we have enough work cut out for us if we want to make this fast as it is.

core/src/main/scala/zio/logging/Logger.scala
```diff
@@ -0,0 +1,64 @@
...
+    ZIO.accessM[Clock with Console](env =>
+      make(
+        logLevel,
+        (context, line) =>
+          (for {
+            date  <- currentDateTime
+            level = context.get(LogAnnotation.Level)
+            loggerName = context.get(LogAnnotation.Name) match {
+              case Nil => classNameForLambda(line).getOrElse("ZIO.defaultLogger")
+              case _   => LogAnnotation.Name.render(context.get(LogAnnotation.Name))
```
jdegoes: I guess we only pay the cost if there is no name.

jdegoes: Change to:

```scala
case list => LogAnnotation.render(list)
```

to avoid getting it twice.

core/src/main/scala/zio/logging/LoggerLike.scala
```diff
@@ -0,0 +1,63 @@
+package zio.logging
+import zio.{ UIO, ZIO }
+
+trait LoggerLike[-A] { self =>
```
jdegoes: You missed a few improvements like `decorate` and so forth. I'd copy them over, they won't break anything else.

core/src/main/scala/zio/logging/package.scala
```diff
@@ -0,0 +1,14 @@
+package zio
+
+package object logging {
+  def log(line: => String): ZIO[Logger, Nothing, Unit] =
```
jdegoes: If you want to ship this right away, then rather than putting Logger in the environment, I'd make a trait `Logging` that has `def logging: Logger`. That way it can compose with other environments. Then all these can use `ZIO[Logging, ...]`.

The problem with putting `Logger` in the environment is that it can't compose easily (or at all) with other environments.

core/src/main/scala/zio/logging/package.scala
```diff
@@ -0,0 +1,14 @@
...
+    ZIO.accessM[Logger](_.log(line))
+
+  def log(level: LogLevel)(line: => String): ZIO[Logger, Nothing, Unit] =
```
jdegoes: This only has 3 helper functions. If you trait `Logging`, you can put _all_ helper functions in there, and delegate these three (the very common ones) to those.

That way, importing `zio.logging._` only gives you these 3, but you can access the other 8 or so by using `Logging.XYZ`.

Let me know if this doesn't make sense!

core/src/main/scala/zio/logging/Logger.scala
pshemass: I added this as an option, sometimes useful but I have mix feeling about this as well. it's used only when ```LogAnnotation.Name``` is empty.

core/src/main/scala/zio/logging/Logger.scala
pshemass: I don't like that because it's confusing. I added this because when you start default level was ```Off``` I didn't see logs I thought something wrong with my setup. I thing users might end up in the same situation. I changed neutral value in ```LogAnnotation.Level``` to be ```Info``` then user will see logs and everyone will be happy.

core/src/main/scala/zio/logging/Logger.scala
pshemass: i chose ```ignore```

core/src/main/scala/zio/logging/Logger.scala
pshemass: done

core/src/main/scala/zio/logging/Logger.scala
pshemass: done

core/src/main/scala/zio/logging/LoggerLike.scala
pshemass: added derive

core/src/main/scala/zio/logging/package.scala
pshemass: ```Logging``` trait added with all helpers

core/src/main/scala/zio/logging/package.scala
pshemass: all helpers in ```Logging```

core/src/main/scala/zio/logging/Logger.scala
pshemass: done

core/src/main/scala/zio/logging/Logger.scala
pshemass: let me know if you want to keep it.

core/src/main/scala/zio/logging/LogAnnotation.scala
```diff
@@ -0,0 +1,37 @@
...
+
+  /**
+   * The `Level` annotation keeps track of log levels.
+   */
+  val Level = LogAnnotation[LogLevel]("level", LogLevel.Info, (_, r) => r, _.render)
+
+  /**
+   * The `Name` annotation keeps track of logger names.
+   */
+  val Name = LogAnnotation[List[String]]("name", Nil, _ ++ _, _.mkString("."))
```
jdegoes: Could be useful to add `CorrelationId` as `UUID`, with a combine that chooses the last, and a renderer that prints out the hex dash encoding (?).

core/src/main/scala/zio/logging/Logger.scala
jdegoes: I think I would prefer _not_ to keep it unless users ask for it... they can always do this themselves with a thin wrapper.

core/src/main/scala/zio/logging/LoggerLike.scala
```diff
@@ -0,0 +1,77 @@
...
+   * Logs the specified element at the specified level. Implementations may
+   * override this for greater efficiency.
+   */
+  def log(level: LogLevel)(line: => A): URIO[R, Unit] =
+    locally(_.annotate(LogAnnotation.Level, level))(log(line))
+
+  /**
+   * Produces a named logger.
+   */
+  def named(name: String): LoggerLike[A, R] =
```
jdegoes: This implementation should be replaced by the simpler one in my Issue, which uses `derive`.

core/src/main/scala/zio/logging/Logging.scala
```diff
@@ -1,21 +1,72 @@
...
+  def make[R](logger: (LogContext, => String) => URIO[R, Unit]) =
+    Logger
+      .make(logger)
+      .map(l =>
+        new Logging[R] {
+          override def logger: Logger[R] = l
+        }
+      )
+
+  val ignore: UIO[Logging[Any]] =
```
jdegoes: Like it. 👍

core/src/main/scala/zio/logging/Logging.scala
```diff
@@ -1,21 +1,72 @@
...
+    make((_, _) => ZIO.unit)
+
+  def console(
+    format: (LogContext, => String) => String
+  ): URIO[Console with Clock, Logging[Console with Clock]] =
```
jdegoes: Good simplification. 👍

core/src/main/scala/zio/logging/Logging.scala
```diff
@@ -1,21 +1,72 @@
...
+  def logger: Logger[R]
 
-trait Logging[Message] {
-  def logging: Logging.Service[Any, Message]
 }
 
 object Logging {
 
-  trait Service[-R, Message] {
+  def make[R](logger: (LogContext, => String) => URIO[R, Unit]) =
```
jdegoes: I'd add the return type here because it's useful documentation on top-level functions.

core/src/main/scala/zio/logging/Logging.scala
```diff
@@ -1,21 +1,72 @@
...
+
+  def log(level: LogLevel)(line: => String): ZIO[Logging[Any], Nothing, Unit] =
+    ZIO.accessM[Logging[Any]](_.logger.log(level)(line))
+
+  def locallyAnnotate[A, R <: Logging[Any], E, A1](annotation: LogAnnotation[A], value: A)(
+    zio: ZIO[R, E, A1]
+  ): ZIO[Logging[Any] with R, E, A1] =
+    ZIO.accessM[Logging[Any] with R](_.logger.locallyAnnotate[A, R, E, A1](annotation, value)(zio))
+
+  def contramap[A1](f: A1 => String): URIO[Logging[Any], LoggerLike[A1, Any]] =
```
jdegoes: This doesn't feel right here. I think I'd delete it.

core/src/main/scala/zio/logging/Logging.scala
```diff
@@ -1,21 +1,72 @@
...
+    ZIO.access[Logging[Any]](_.logger.contramap(f))
+
+  def derive(f: LogContext => LogContext): URIO[Logging[Any], LoggerLike[String, Any]] =
+    ZIO.access[Logging[Any]](_.logger.derive(f))
```
jdegoes: I'd delete `derive`, too.

core/src/main/scala/zio/logging/Logging.scala
pshemass: I had the same feeling. if some wanna do it could still use ```logger``` method and then ```contramap```

core/src/main/scala/zio/logging/Logger.scala
```diff
@@ -0,0 +1,41 @@
...
+import zio.internal.stacktracer.Tracer
+import zio.internal.stacktracer.ZTraceElement.{ NoLocation, SourceLocation }
+import zio.internal.stacktracer.impl.AkkaLineNumbersTracer
+import zio.internal.tracing.TracingConfig
+import zio.{ FiberRef, UIO, URIO, ZIO }
+
+/**
+ * A logger of strings.
+ */
+trait Logger[-R] extends LoggerLike[String, R]
```
jdegoes: Let's delete the `R` parameter from `Logger` as it won't be supported in RC18 and later.

core/src/main/scala/zio/logging/LoggerLike.scala
```diff
@@ -0,0 +1,77 @@
+package zio.logging
+import zio.{ UIO, URIO, ZIO }
+
+trait LoggerLike[-A, -R] { self =>
```
jdegoes: Let's delete the `R` parameter from `LoggerLike`.

core/src/main/scala/zio/logging/Logging.scala
```diff
@@ -1,21 +1,72 @@
...
+    ZIO.accessM[Logging[Any]](_.logger.logContext)
 
-    def trace(message: => Message): ZIO[R, Nothing, Unit]
-    def debug(message: => Message): ZIO[R, Nothing, Unit]
-    def info(message: => Message): ZIO[R, Nothing, Unit]
-    def warning(message: => Message): ZIO[R, Nothing, Unit]
-    def error(message: => Message): ZIO[R, Nothing, Unit]
-    def error(message: => Message, cause: Cause[Any]): ZIO[R, Nothing, Unit]
-  }
+  def named(name: String): URIO[Logging[Any], LoggerLike[String, Any]] =
```
jdegoes: I'd delete this one too. Essentially any Logger function that returns a new Logger feels more natural to express on the Logger itself, and not as a helper function inside the companion object.

core/src/main/scala/zio/logging/Logger.scala
pshemass: sure I will move this to slf4j and hide it.

core/src/main/scala/zio/logging/LoggerLike.scala
```diff
@@ -0,0 +1,77 @@
...
+
+      def log(line: => A): URIO[R, Unit] = locally(f)(self.log(line))
+
+      def logContext: UIO[LogContext] = self.logContext
+    }
+
+  /**
+   * Modifies the log context in the scope of the specified effect.
+   */
+  def locally[R1, E, A1](f: LogContext => LogContext)(zio: ZIO[R1, E, A1]): ZIO[R1, E, A1]
```
jdegoes: We're going to need to add the following variant:

```scala
def locallyM[R1, E, A1](f: LogContext => URIO[R1, LogContext])(zio: ZIO[R1, E, A1]): ZIO[R1, E, 
A1]
```

And support in the implementations.

This will allow us to introduce `Timed` and other `LogAnnotation` types that are effectful.

core/src/main/scala/zio/logging/LogAnnotation.scala
pshemass: for correlationId it could be ```Option[UUID]``` because there are part of program that may run without correlation id e.g. startup logs, warmup, periodic log. Of course it depends how do we think about correlation id. what do you think?

core/src/main/scala/zio/logging/LoggerLike.scala
pshemass: on my way

core/src/main/scala/zio/logging/Logging.scala
pshemass: of course

core/src/main/scala/zio/logging/LogAnnotation.scala
jdegoes: Yes, of course, `Option[UUID]`.

core/src/main/scala/zio/logging/Logger.scala
pshemass: I added to avoid usage of ```provide``` in ```make[R]``` but I agree it was silly reason.

core/src/main/scala/zio/logging/LoggerLike.scala
pshemass: I thinking about this and I think ```locallyM``` will be useful but i don't know about ```Timed``` annotation. Realistically speaking nobody define this on logging line level or even block of the code this is typically on application level. We probably should think about expressing ```LogFormat``` in some convient way. 

I added ```locallyM```

core/src/main/scala/zio/logging/Logging.scala

PR #41, opened 2020-02-10
Miscellaneous

pshemass:


core/src/main/scala/zio/logging/Logging.scala
```diff
@@ -36,6 +37,12 @@ object Logging {
       } yield ()
     )
 
+  def error(cause: Cause[Any]): ZIO[Logging, Nothing, Unit] =
+    log(LogLevel.Error)(cause.prettyPrint)
+
+  def throwable(t: Throwable): ZIO[Logging, Nothing, Unit] =
```
jdegoes: I would actually implement this one by wrapping the `Throwable` into a `Cause`, and then delegating to `error`.

core/src/main/scala/zio/logging/Logging.scala
```diff
@@ -36,6 +37,12 @@ object Logging {
       } yield ()
     )
 
+  def error(cause: Cause[Any]): ZIO[Logging, Nothing, Unit] =
```
jdegoes: Would you mind sorting these methods alphabetically, here and in subclasses? 🙏

core/src/main/scala/zio/logging/package.scala
```diff
@@ -7,6 +7,15 @@ package object logging {
   def log(level: LogLevel)(line: => String): ZIO[Logging, Nothing, Unit] =
     Logging.log(level)(line)
 
+  def error(cause: Cause[Any]): ZIO[Logging, Nothing, Unit] =
```
jdegoes: To avoid needless name collisions, maybe we should prefix all "package object methods" with log, e.g. `logError`, `logThrowable`, `logLocally`, etc.

core/src/main/scala/zio/logging/LogAnnotation.scala
```diff
@@ -8,6 +8,9 @@ import scala.reflect.ClassTag
  */
 final case class LogAnnotation[A: ClassTag](name: String, initialValue: A, combine: (A, A) => A, render: A => String) {
   self =>
+
+  def apply(value: A): LogContext => LogContext = _.annotate(self, combine(initialValue, value))
```
jdegoes: Would be nice to use this in some code to do cleanup. There are a few places that could benefit.

core/src/main/scala/zio/logging/Logging.scala

core/src/main/scala/zio/logging/Logging.scala

core/src/main/scala/zio/logging/package.scala
pshemass: done

core/src/main/scala/zio/logging/LogAnnotation.scala
pshemass: I actually dropped `locallyAnnotate` because with this I don't think we need it. I updated examples and docs accordingly.

PR #52, opened 2020-03-04
ZIO upgrade to RC18

pshemass:


core/src/main/scala/zio/logging/Logging.scala
```diff
@@ -27,32 +20,29 @@ object Logging {
   def error(cause: Cause[Any]): ZIO[Logging, Nothing, Unit] =
     log(LogLevel.Error)(cause.prettyPrint)
 
-  val ignore: UIO[Logging] =
+  val ignore: ZLayer[Any, Nothing, Logging] =
```
adamgfraser: Very minor but could be slightly more idiomatically `ZLayer.NoDeps[Nothing, Logging]`.

core/src/main/scala/zio/logging/Logging.scala
```diff
@@ -27,32 +20,29 @@ object Logging {
...
     make((_, _) => ZIO.unit)
 
   def locally[R1 <: Logging, E, A1](f: LogContext => LogContext)(zio: ZIO[R1, E, A1]): ZIO[R1, E, A1] =
-    ZIO.accessM[R1](_.logger.locally(f)(zio))
+    ZIO.accessM[R1](_.get.locally(f)(zio))
```
adamgfraser: I think you can actually omit the type annotation here and below since it can be inferred from the expected return type of the method.

core/src/test/scala/zio/logging/LoggerSpec.scala
```diff
@@ -4,159 +4,151 @@ import java.time.OffsetDateTime
...
+object LoggerSpec extends DefaultRunnableSpec {
 
-      case class TestLogger(ref: Ref[Vector[(LogContext, String)]], logger: Logger) extends Logger {
-        override def locally[R, E, A1](f: LogContext => LogContext)(zio: ZIO[R, E, A1]): ZIO[R, E, A1] =
-          logger.locally(f)(zio)
-        override def log(line: => String): UIO[Unit] =
-          logger.log(line)
-        override def logContext: UIO[LogContext] =
-          logger.logContext
+  case class TestLogger(ref: Ref[Vector[(LogContext, String)]], logger: Logger) extends Logger {
```
adamgfraser: You could potentially turn this into a layer. Then you could provide a copy to each of your tests by just calling `provideCustomLayer` once on the whole suite. Then in your tests you could just call the methods you want with the environmental accessors and verify that you get the expected result without having to create the test logger within each tes.

core/src/main/scala/zio/logging/Logging.scala
pshemass: thanks I was looking for that :)

core/src/main/scala/zio/logging/Logging.scala
pshemass: I will try

PR #61, opened 2020-03-15
Adding scala.js support and two loggers

aparo:
This pull request add scala.js support to zio-logging and two new loggers:
- js console logger to use standard browser console to print logs
- Ajax POST backend to send event to a backend.

I added documentation for both loggers and test for console one.


pshemass:
@aparo this all looks great and I really appreciate your PR! 

before we can merge this you need to fix formatting by running `sbt fmt` in root of your local project. 

other think we need to make sure that test are run and artifacts are properly published for scala.js. 
@mijicd could you please help us here with setting up tests and publishing artifact in circleCi config?

mijicd:
@pshemass Release is already there, although it uses the old Sonatype context (I'll submit a PR for that). Regarding JS steps, safest way is to just copy relevant steps from ZIO config and add it to `config.yml`.

pshemass:
> @pshemass Release is already there, although it uses the old Sonatype context (I'll submit a PR for that). Regarding JS steps, safest way is to just copy relevant steps from ZIO config and add it to `config.yml`.

@aparo would you mind to adjust `config.yml`?

aparo:
Updated with code formatting

pshemass:
> Updated with code formatting

@aparo could you please look into failing tests?

pshemass:
@aparo I eventually managed to fix the build. Thanks for contribution!

PR #65, opened 2020-03-18
package object replaced with log object

pshemass:


core/src/main/scala/zio/logging/log.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  def locally[A, R <: Logging, E, A1](fn: LogContext => LogContext)(zio: ZIO[R, E, A1]): ZIO[Logging with R, E, A1] =
+    ZIO.accessM(_.get.logger.locally(fn)(zio))
+
+  def locallyM[A, R <: Logging, E, A1](
+    fn: LogContext => URIO[R, LogContext]
+  )(zio: ZIO[R, E, A1]): ZIO[Logging with R, E, A1] =
+    ZIO.accessM(_.get.logger.locallyM(fn)(zio))
+
+  def logger: URIO[Logging, Logger] =
```
jdegoes: Could be `val`

jdegoes:
Nice work!

PR #66, opened 2020-03-19
Use the Throwable version of slf4j logger methods

PeterHajdu:
Hi

I'm not sure if I understood the library structure correctly but this seemed to be a possible solution.  What do you think?


pshemass:
@PeterHajdu this is really good idea :)

PR #78, opened 2020-03-31
reorg towards layers

pshemass:


core/src/main/scala/zio/logging/log.scala
```diff
@@ -5,44 +5,41 @@ import zio.{ Cause, URIO, ZIO }
 
 object log {
   def apply(line: => String): ZIO[Logging, Nothing, Unit] =
-    ZIO.accessM[Logging](_.get.logger.log(line))
+    ZIO.accessM[Logging](_.get.log(line))
```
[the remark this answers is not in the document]
pshemass: yep

docs/overview/index.md
```diff
@@ -92,47 +158,45 @@ Library provides object `log` that exposes basic methods:
 ```scala
 object log {
   def apply(line: => String): ZIO[Logging, Nothing, Unit] =
-    ZIO.accessM[Logging](_.get.logger.log(line))
+    ZIO.accessM[Logging](_.get.log(line))
```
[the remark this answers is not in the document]
pshemass: you are right. I will remove this.

core/src/main/scala/zio/logging/Logging.scala
```diff
@@ -1,45 +1,95 @@
...
+  val context: URIO[Logging, LogContext] =
+    ZIO.accessM[Logging](_.get.logContext)
+
+  def debug(line: => String): ZIO[Logging, Nothing, Unit] =
+    ZIO.accessM[Logging](_.get.debug(line))
+
+  def error(line: => String): ZIO[Logging, Nothing, Unit] =
+    ZIO.accessM[Logging](_.get.error(line))
+
+  def error(line: => String, cause: Cause[Any]): ZIO[Logging, Nothing, Unit] =
```
mschuwalow, line 40: What do you think about reverting to the logError naming scheme? I think it will be quite common for people to do a wildcard import zio.logging.Logging._ (would look nicer if these were in the package object imo), but with the current naming this will probably cause name clashes.

The alternative of having to log everywhere with Logging.info("bla") seems too verbose

core/src/main/scala/zio/logging/Logging.scala
```diff
@@ -1,45 +1,95 @@
...
 import zio.console.{ putStrLn, Console }
-import zio._
 
 object Logging {
-  type Logging = Has[Logging.Service]
-
-  trait Service {
-    def logger: Logger
-  }
+  type Logging = Has[Logger[String]]
```
mschuwalow, line 8: the import for this looks very awkward imo (import zio.logging.Logging.Logging) and is required everywhere

core/src/main/scala/zio/logging/Logging.scala
```diff
@@ -1,45 +1,95 @@
...
+            .getOrElse("")
+          _ <- putStrLn(date.toString + " " + level + " " + loggerName + " " + format(context, line) + " " + maybeError)
+        } yield (),
+      rootLoggerName
     )
 
+  val context: URIO[Logging, LogContext] =
+    ZIO.accessM[Logging](_.get.logContext)
+
+  def debug(line: => String): ZIO[Logging, Nothing, Unit] =
```
mschuwalow, line 34: I think these can be polymorphic without any loss in ergonomics for the String case

core/src/main/scala/zio/logging/Logging.scala
pshemass: this is good idea. I will keep `log.debug` and here we will have `logDebug`. @jdegoes what do you think?

core/src/main/scala/zio/logging/Logging.scala
pshemass: yes I know... what do you recommend? package object with this type?

core/src/main/scala/zio/logging/Logging.scala
pshemass: I will keep this for now. We need to think about "Structured Logging" probably via `Message` from LogStage. then this will be probably polymorphic...

core/src/main/scala/zio/logging/Logging.scala
mschuwalow: I think that would be nicer, yeah :)

core/src/main/scala/zio/logging/log.scala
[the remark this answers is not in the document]
jdegoes: Then how about also implementing it like `val log = Logging`. This way you can eliminate the duplication.

core/src/main/scala/zio/logging/Logging.scala
jdegoes: I don't think someone will import `Logging._` because it contains all the constructors.

Maybe keeping the `log` helper is sufficient, e.g. `log.info` reads rather well (in my eyes), at least as well as `logInfo`.

core/src/main/scala/zio/logging/Logging.scala
jdegoes: Agreed, `Logging` should be in package object like ZIO core convention.

core/src/main/scala/zio/logging/Logging.scala
jdegoes: I think for LogStage, we can have:

```scala
package object logstage {
  type LogStage = Has[Logger[Message]]
  object LogStage {
    // helpers
  }
}
```

core/src/main/scala/zio/logging/log.scala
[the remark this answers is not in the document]
neko-kai: `val log` will not be auto-imported in IDEs, I'd keep duplication just for the UX

core/src/main/scala/zio/logging/Logging.scala
neko-kai: there's already full layer support in logstage - https://izumi.7mind.io/latest/release/doc/logstage/index.html#zio-environment-support
It's a matter of just re-exporting if you want to add it here, I guess

core/src/main/scala/zio/logging/log.scala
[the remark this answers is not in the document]
pshemass: I'm going to keep both ```log``` and ```Logging``` implementations and ```log``` is using ```Logging``` to avoid duplicates.

core/src/main/scala/zio/logging/Logging.scala
pshemass: in my opinion ```log.info``` looks better. I'm not going to change this for now

core/src/main/scala/zio/logging/Logging.scala
pshemass: done

core/src/main/scala/zio/logging/Logging.scala
pshemass: thanks @neko-kai I will open separate PR for that.

PR #84, opened 2020-04-18
Add slf4j with annotations on MDC

pierangeloc:
Added an implementation of slf4j logger that puts a defined set of annotations from the LogContext into MDC logging context


pierangeloc:
This is an attempt to satisfy this

pierangeloc:
I see there are conflicts (master was updated recently). I will fix them and open a different PR

PR #85, opened 2020-04-18
slf4j logger with annotations on mdc context

pierangeloc:
Added an implementation of slf4j logger that puts a defined set of annotations from the LogContext into MDC logging context.

### DIFF
--- a/slf4j/src/main/scala/zio/logging/slf4j/Slf4jLogger.scala
+++ b/slf4j/src/main/scala/zio/logging/slf4j/Slf4jLogger.scala
@@ -1,6 +1,6 @@
 package zio.logging.slf4j
 
-import org.slf4j.LoggerFactory
+import org.slf4j.{ LoggerFactory, MDC }
 import zio.internal.Tracing
 import zio.internal.stacktracer.Tracer
 import zio.internal.stacktracer.ZTraceElement.{ NoLocation, SourceLocation }
@@ -9,6 +9,7 @@ import zio.internal.tracing.TracingConfig
 import zio.logging.Logging
 import zio.logging._
 import zio.{ ZIO, ZLayer }
+import scala.collection.JavaConverters._
 
 object Slf4jLogger {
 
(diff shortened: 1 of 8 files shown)


pshemass:
@pierangeloc looks good to me please add example to documentation and test if possible.

slf4j/src/main/scala/zio/logging/slf4j/Slf4jLogger.scala
```diff
@@ -53,4 +54,40 @@ object Slf4jLogger {
       },
       rootLoggerName = rootLoggerName
     )
+
+  def makeWithAnnotationsAsMdc(
```
pshemass, line 58: please add example to documentation and test if possible

pierangeloc: @pshemass sure, will do that 👍

pshemass: @pierangeloc any update on that? do you need help?

pierangeloc: @pshemass sorry I got dragged by other tasks during the week. I'm done with testing, need just to update docs. I should be done with that today/tomorrow

pierangeloc: @pshemass added test and docs, can you have a look?

sebver:
Hi @pierangeloc, I'm implementing something like this myself now and stumbled upon this PR. The code works for me, but I noticed it doesn't take into account the initial value. Is this something we want to do?

Not the best code probably but this way it takes into account:
```
val mdc: Map[String, String] =
    mdcAnnotations
      .asInstanceOf[List[logging.LogAnnotation[Any]]]
      .map(k => k.name -> renderedContext.getOrElse(k.name, k.render(k.initialValue)))
      .toMap
```

pshemass:
@sebver feel free to open PR for your change

PR #95, opened 2020-05-14
reformat with scalafmt 2.5.2

pshemass:


core/src/main/scala/zio/logging/LogLevel.scala
```diff
@@ -16,7 +16,8 @@ sealed trait LogLevel { self =>
 
   def min(that: LogLevel): LogLevel = if (self > that) that else self
 }
-object LogLevel {
+object LogLevel       {
```
pshemass, line 19: @mijicd do you know how to get rid of this?

mijicd: Yes, you need to wrap it like this:

```scala
// format: off
object Dummy {
// format: on
```

pshemass: this is disgusting :(

mijicd: I'm not sure what's uglier, the formatting or these "magic" comments.

pshemass: I agree it's pretty hard to judge :D

PR #111, opened 2020-06-08
Lazy Logging

tusharmath:
## Feature
A lot of times, an effect should be executed only for logging purposes. This can cause performance issues even when the logging is turned off since the effect will still get evaluated.

## Current
```scala
for {
  f <- foo
  b <- bar
  _ <- log.debug(b) // b is needed only for logging purposes
} yield f
```

## Proposed
```scala
for {
  f <- foo
  _ <- log.debug(bar) // bar get's evaluated only when LogLevel is "debug"
} yield f
```

### DIFF
--- a/core/src/main/scala/zio/logging/Logger.scala
+++ b/core/src/main/scala/zio/logging/Logger.scala
@@ -24,10 +24,20 @@ trait Logger[-A] { self =>
     self.log(LogLevel.Debug)(line)
 
   /**
+   * Evaluates the specified element based on the LogLevel set and logs at the debug level
+   */
+  def debug(line: UIO[A]): UIO[Unit] = line >>= debug
+
+  /**
    * Logs the specified element at the error level.
    */
   def error(line: => A): UIO[Unit] =
     self.log(LogLevel.Error)(line)
+
+  /**
+   * Evaluates the specified element based on the LogLevel set and logs at the error level
+   */
+  def error(line: UIO[A]): UIO[Unit] = line >>= error
 
   /**
    * Logs the specified element at the error level with cause.
@@ -36,6 +46,11 @@ trait Logger[-A] { self =>
     self.locally(LogAnnotation.Cause(Some(cause))) {
       self.log(LogLevel.Error)(line)
     }
+
+  /**
+   * Evaluates the specified element based on the LogLevel set and logs at the error level
+   */
+  def error(line: UIO[A], cause: Cause[Any]): UIO[Unit] = line.flatMap(l => error(l, cause))
 
   /**
    * Derives a new logger from this one, by applying the specified decorator
@@ -56,6 +71,11 @@ trait Logger[-A] { self =>
    */
   def info(line: => A): UIO[Unit] =
     self.log(LogLevel.Info)(line)
+
+  /**
+   * Evaluates the specified element based on the LogLevel set and logs at the info level
+   */
+  def info(line: UIO[A]): UIO[Unit] = line >>= info
 
   /**
    * Modifies the log context in the scope of the specified effect.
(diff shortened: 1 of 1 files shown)


pshemass:
I'm not sure about that. @jdegoes what do you think?

core/src/main/scala/zio/logging/Logger.scala
```diff
@@ -23,12 +23,22 @@ trait Logger[-A] { self =>
...
   /**
    * Logs the specified element at the error level.
    */
   def error(line: => A): UIO[Unit] =
     self.log(LogLevel.Error)(line)
 
+  /**
+   * Evaluates the specified element based on the LogLevel set and logs at the error level
+   */
+  def error(line: UIO[A]): UIO[Unit] = line >>= error
```
jdegoes, line 40: This should be called `errorM`, by convention.

core/src/main/scala/zio/logging/Logger.scala
```diff
@@ -37,6 +47,11 @@ trait Logger[-A] { self =>
       self.log(LogLevel.Error)(line)
     }
 
+  /**
+   * Evaluates the specified element based on the LogLevel set and logs at the error level
+   */
+  def error(line: UIO[A], cause: Cause[Any]): UIO[Unit] = line.flatMap(l => error(l, cause))
```
jdegoes, line 53: This should be called `errorM`, by convention.

core/src/main/scala/zio/logging/Logger.scala
```diff
@@ -57,6 +72,11 @@ trait Logger[-A] { self =>
   def info(line: => A): UIO[Unit] =
     self.log(LogLevel.Info)(line)
 
+  /**
+   * Evaluates the specified element based on the LogLevel set and logs at the info level
+   */
+  def info(line: UIO[A]): UIO[Unit] = line >>= info
```
jdegoes, line 78: This should be called `infoM`, by convention.

core/src/main/scala/zio/logging/Logger.scala
```diff
@@ -111,9 +131,19 @@ trait Logger[-A] { self =>
   def trace(line: => A): UIO[Unit] =
     self.log(LogLevel.Trace)(line)
 
+  /**
+   * Evaluates the specified element based on the LogLevel set and logs at the trace level
+   */
+  def trace(line: UIO[A]): UIO[Unit] = line >>= trace
```
jdegoes, line 137: This should be called `traceM`, by convention.

core/src/main/scala/zio/logging/Logger.scala
```diff
@@ -111,9 +131,19 @@ trait Logger[-A] { self =>
...
   /**
    * Logs the specified element at the warn level.
    */
   def warn(line: => A) =
     self.log(LogLevel.Warn)(line)
+
+  /**
+   * Evaluates the specified element based on the LogLevel set and logs at the warn level
+   */
+  def warn(line: UIO[A]): UIO[Unit] = line >>= warn
```
jdegoes, line 148: This should be called `warnM`, by convention.

jdegoes:
With slight modifications, looks great to me! I left some comments. Thanks @tusharmath!

pshemass:
> With slight modifications, looks great to me! I left some comments. Thanks @tusharmath!

maybe I don't see something but @tusharmath proposed "Lazy Logging" - based on his comment `bar` should be evaluated only when `LogLevel.Debug` is set. This is not the case based on his code. I agree with naming convention though :)

@tusharmath could you please add test for this?

tusharmath:
@pshemass Yes exactly. I know the PR doesn't actually do what it says. I guess, this requires a lot more deeper integration with the loggers.

pshemass:
@tusharmath don't worry :) after all if you change those methods names as @jdegoes suggested this code is still useful and can be merged.

PR #139, opened 2020-08-19
introduce LogAppender

pshemass:
Previous design was a little bit oversimplified and ignore more advanced usecases for logger that could be created without use of Slf4j.

Proposition is to introduce `LogFormat` and `LogAppender` as typical frameworks for logging provide. 
`LogFormat`  represents function from `(LogContext, A) => String`. In `LogFormat` we can add timestamp logLevel or other informations from `LogContext`. 
`LogAppender` is function that persists formatted line to storage of choice. Console, Console with errors on StdErr, File are provided by library. Probably we could add Async appender.

Example of use
```
ZLayer.requires[Console with Clock] ++ 
LogAppender.consoleErr(LogLevel.Info, LogFormat.ColoredLogFormat((_, s) => s)) >>> 
Logging.makeWithTimestamp(Some("myApp"))
```
 
another example of new API
```
Logging.consoleErr(
  level = LogLevel.Info, 
  format = LogFormat.ColoredLogFormat()
) >>> Logging.withRootLoggerName("my-component")
```


pshemass:
@mijicd sorry I bother you again but it looks like build is broken https://app.circleci.com/pipelines/github/zio/zio-logging/356/workflows/1bf47caf-6fad-4549-9c7f-119095e43e9a/jobs/1752 please help

mijicd:
@pshemass retriggered, it's showing the attitude :)

pshemass:
@softinio @mijicd sorry for silly question but for 2.11 it seems that I cannot define `type LogAppender[A] = Has[LogAppender.Service[A]]` in package object `logging` because it causes circular dependency. is this known issue or I'm doing something stupid?

examples/src/main/scala/zio/logging/Examples.scala
```diff
@@ -13,19 +13,20 @@ object Examples extends zio.App {
...
     Logging.console(format =
       LogFormat.ColoredLogFormat((context, message) => logFormat.format(context(correlationId), message))
-    )
+    ) >>> Logging.modifyLogger(_.derive(correlationId("0000"))) >>> Logging.withRootLoggerName("root-logger")
 
+  import zio.duration._
   override def run(args: List[String]) =
     (for {
       fiber <- log.locally(correlationId("1234"))(ZIO.unit).fork
       _     <- log.info("info message without correlation id")
```
leszekgruchala, line 25: Is this still without a correlation ID? In the line 19 you modify a logger and I would assume it will initially have correlationId 0000 applied.

core/src/main/scala/zio/logging/LogAppender.scala
```diff
@@ -0,0 +1,80 @@
...
+  def file[A: Tag](filename: String, format0: LogFormat[A]) =
+    ZManaged.makeEffect {
+      val path = Paths.get(filename)
+      Files.newBufferedWriter(path)
+    }(_.close())
+      .map(writer =>
+        new Service[A] {
+
+          override def write(ctx: LogContext, msg: => A): UIO[Unit] =
+            ZIO.effectTotal(writer.write(format0.format(ctx, msg)))
```
leszekgruchala, line 69: Why `.effectTotal` if file writer can fail with some IO exception?

core/src/test/scala/zio/logging/LoggerSpec.scala
```diff
@@ -92,6 +94,24 @@ object LoggerSpec extends DefaultRunnableSpec {
             )
           )
       },
+      testM("derive") {
+        val counter = LogAnnotation[Int](
+          "counter",
```
leszekgruchala, line 99: Please name these parameters, it will be easier to read.

examples/src/main/scala/zio/logging/Examples.scala
pshemass: yes, that correct

core/src/main/scala/zio/logging/LogAppender.scala
pshemass: it can indeed. question is what you can do about this? technically when you writing to console or to slf4j you may get exception too. 

that will start discussion if we should include errors in logging interface. we decided to not for the time being.

examples/src/main/scala/zio/logging/Examples.scala
leszekgruchala: What exactly? That there will be correlationID or not?

core/src/main/scala/zio/logging/LogAppender.scala
leszekgruchala: Ok, but if `writer.write` fails the whole fiber will die, correct?

examples/src/main/scala/zio/logging/Examples.scala
pshemass: I see what you mean - message is misleading. in this line you will have correlation id from initial value.

core/src/main/scala/zio/logging/LogAppender.scala
pshemass: we can ignore it but I don't really like sweeping stuff under rug approach. so to answer you question yes it's desired i want program to exit if logging module failed to write log to file.

would you expect different behavior?

examples/src/main/scala/zio/logging/Examples.scala
pshemass: I changed message

core/src/test/scala/zio/logging/LoggerSpec.scala
pshemass: done

core/src/main/scala/zio/logging/LogAppender.scala
leszekgruchala: Yes. I'd expect the library to be safe to use so I don't have to try/catch library calls. We're behind exception driven development. If some library API can fail for some reason, I'd like to know it.

pshemass: for me this will very uncomfortable api that I need to handle error from logging that I can do nothing about. Try to visual code with this error and how would you handle this.

PR #146, opened 2020-09-05
Add async file logging and more features to file appender

reibitto:
I noticed there's a `file` LogAppender and I wanted to use it, but it didn't have a constructor in `Logging`. It looks like it was still a work in progress. Here's my attempt at finishing it, as well as adding an async file logger which does two things:
- Return to the caller as quickly as possible (the "write" is just an enqueuing operation)
- Makes flushing more efficient by batching them (the batch size is configurable)

A few other things:
- `LogAppender.file` wasn't inserting a newline at the end unlike `LogAppender.console`. This means the output would be on a single line. And since `BufferedWriter` flushes on newline, flushing wasn't happening properly.
- I added LogLevel filtering to both `file` and `fileAsync`
- I changed the filename param to be a `Path` to be more consistent with ZIO (like ZStream), and because it makes it more obvious that it's a path and not necessarily just a filename. And specific types are generally better than String for catching mistakes earlier.

I didn't quite address one other issue that was bothering me because I figured a discussion about it would be better before making any changes... but the usage of `ZIO.effectTotal` for file writes doesn't seem desirable. I understand the decision came from not wanting to make `LogAppender.write` deal with failures, but having a Logger fail with defects is incredibly difficult to handle. If I don't want crashes due to logging I basically have sandbox every logging statement, which means managing my own class full of helper methods. It would duplicate a lot of what the `Logging` object does.

I think there are other ways to handle that. Perhaps an ADT which specifies the error handling strategy, like: `Die`, `Ignore`, and `Warn`.

I was thinking `Warn` would be something like: if an exception happens during file writing, it would print a warning to `System.err` and then give up with file I/O without crashing the program.

### DIFF
--- a/core/src/main/scala/zio/logging/LogAppender.scala
+++ b/core/src/main/scala/zio/logging/LogAppender.scala
@@ -1,11 +1,11 @@
 package zio.logging
 
-import java.io.FileWriter
-import java.nio.file.{ Files, Path }
+import java.nio.charset.Charset
+import java.nio.file.Path
 
 import izumi.reflect.Tag
 import zio.console._
-import zio.{ Has, Task, UIO, ULayer, URIO, ZIO, ZLayer, ZManaged, ZQueue }
+import zio.{ Has, UIO, ULayer, URIO, ZIO, ZLayer, ZManaged, ZQueue }
 
 /**
  * Represents log writer function that turns A into String and put in console or save to file.
@@ -40,6 +40,23 @@ object LogAppender {
       )
       .toLayer
 
+  def async[A: Tag](logEntryBufferSize: Int): ZLayer[Appender[A], Nothing, Appender[A]] = {
+    case class LogEntry(ctx: LogContext, line: () => A)
+    ZManaged.accessManaged[Appender[A]](env =>
+      ZQueue
+        .bounded[LogEntry](logEntryBufferSize)
+        .tap(queue => queue.take.flatMap(entry => env.get.write(entry.ctx, entry.line())).forever.forkDaemon)
+        .toManaged(_.shutdown)
+        .map(queue =>
+          new Service[A] {
+
+            override def write(ctx: LogContext, msg: => A): UIO[Unit] =
+              queue.offer(LogEntry(ctx, () => msg)).unit
+          }
+        )
+    )
+  }.toLayer
+
   def console[A: Tag](logLevel: LogLevel, format: LogFormat[A]): ZLayer[Console, Nothing, Appender[A]] =
     make[Console, A](format, (_, line) => putStrLn(line)).map(appender =>
       Has(appender.get.filter((ctx, _) => ctx.get(LogAnnotation.Level) >= logLevel))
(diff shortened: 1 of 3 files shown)


core/src/main/scala/zio/logging/LogAppender.scala
```diff
@@ -57,21 +55,49 @@ object LogAppender {
...
-            ZIO.effectTotal(writer.write(format0.format(ctx, msg)))
+            ZIO.effectTotal {
+              writer.write(s"${format0.format(ctx, msg)}${System.lineSeparator}")
+            }
         }
       )
       .toLayer
 
-  def ignore[A: Tag] =
+  def fileAsync[A: Tag](
```
pshemass, line 71: I was thinking that `async` should be a functionality that you can add on top of any appender.

```
LogAppender.file(...) >>> LogAppender.async
```
and 
```
LogAppender.console(...) >>> LogAppender.async
```

Let me know if you can implement it this way.

core/src/main/scala/zio/logging/LogAppender.scala
```diff
@@ -57,21 +55,49 @@ object LogAppender {
...
+  def file[A: Tag](destination: Path, format0: LogFormat[A]): ZLayer[Any, Throwable, Appender[A]] =
+    ZManaged
+      .makeEffect(Files.newBufferedWriter(destination))(_.close())
       .map(writer =>
         new Service[A] {
-
           override def write(ctx: LogContext, msg: => A): UIO[Unit] =
-            ZIO.effectTotal(writer.write(format0.format(ctx, msg)))
+            ZIO.effectTotal {
+              writer.write(s"${format0.format(ctx, msg)}${System.lineSeparator}")
```
pshemass, line 65: +1

pshemass (changes requested):

pshemass:
@reibitto that for this PR. It's great stuff.

core/src/main/scala/zio/logging/LogAppender.scala
reibitto: Hmm, I agree it would be cool to be able to change any logger to be async easily, but I'm not seeing how it can be done generically like that considering the implementations can differ wildly between sync and async. For example, `file` and `fileAsync` don't even use the same type of Writer.

However, I do see the point that there's an opportunity to factor out much of the logic in `fileAsync` into some sort of abstraction that can be shared. At least half the code in `fileAsync` would be common logic.

Unless you have some other ideas that I'm not considering?

reibitto: That does work to a certain extent, but my concern was about how the implementations need to differ if you want a true performance gain. For example, I only want to flush a batch, not for every `write` call.

The difference is more noticeable in the file appender, but it's true even for the console appender. The default println implementation is `synchronized` and flushes for every line.

I'm not currently seeing a way to accomplish this generically. At least not with the current `Service` trait.

pshemass: what about using `BufferedWriter` like that `new BufferedWriter(new FileWriter(...), sizeOfBufferHere)` ?

reibitto: The rabbit hole is quite deep. There are a bunch of things going on:
- The ZQueue implementation is working off the number of "log entries" while BufferedWriter is working at the level of bytes.
- Setting the BufferedWriter size can lead to slightly surprising behavior because FileWriter (or more accurately StreamEncoder) also has its own internal buffer size. Maybe not really noticeable if you choose a larger buffer size though.
- Not flushing manually and working with number of bytes is a possible alternative, but you can get into a state where 1 log entry is only partially written to a file (with the rest still in a buffer). This might not be a problem, but worth noting I think.
- All these Writers are needlessly synchronized. I don't think that's a major concern, but I wonder if there's an NIO solution or something that's more "ideal" for this kind of usage.

I'm fine with a "good enough" solution for a first implementation and then refining it later some point in the future. I understand not wanting to force a big refactor out of this issue. Or not wanting to having 2 separate implementations for each appender.

reibitto: Writing our own small internal `Writer` (mostly just a wrapper) that is "log entry aware" for use in both sync and async might work? I don't know. Just a thought.

pshemass: I was looking for some NIO stuff but it seems that there is nothing out of the box that will help us. 

I like the idea of creating wrapper writer that will be lines aware. Let try this.

pshemass: we may also count number of the logs and flush when reach limit or periodically flush

reibitto:
Alright, I gave the idea a shot. I created a private `LogWriter` that can be used in various ways. Using log4j as a reference, here are the equivalent concepts:
https://logging.apache.org/log4j/2.x/manual/appenders.html#FileAppender

`immediateFlush = true` -> `autoFlushBatchSize = 1`
`bufferedIO = false` -> `bufferedIOSize = None`
`bufferedIO = true` and `bufferSize = 1024` -> `bufferedIOSize = Some(1024)`

Feel free to rename anything if you don't like the names (or tweak the defaults).

Also, I'm not sure what you want to do about `Logging.fileAsync`. Do you want to remove it? It's defined in terms of the new `LogAppender.async`. I couldn't see a way to create a `Logging.async` (if that even makes sense?) to modify the Logger.

If the settings are getting too big, packaging them up in a case class might be something to consider too. Maybe it's still at a manageable level right now. Not sure.

pshemass (approved):

pshemass:
@reibitto I think it's good to go for the time being. we might hit some code reorg and clean up when rolling file appender will be introduce.

PR #157, opened 2020-09-13
Modify failure mode for file I/O exceptions

reibitto:
This is an idea I brought up in the last PR. Basically the idea is to not die with a defect if logging to a file fails. Instead what it does is print the I/O exception to stderr, and then supresses the exception for each subsequent failure. This is to prevent log spam because if one `write` failed, it's not particularly helpful to know that the 2nd, 3rd, etc. also failed with the same exception.

This is just an alternate idea to terminating with ZIO.Die. I could see there being a case for making different configurable modes of failure (like `Die`, `Warn`, `Ignore`). And I don't have strong opinions about what the default should be. As long an option is there, it would satisfy what I'm trying to do.

For an additional reference, log4j seems to have two modes with the `ignoreExceptions` flag.

### DIFF
--- a/core/src/main/scala/zio/logging/LogAppender.scala
+++ b/core/src/main/scala/zio/logging/LogAppender.scala
@@ -2,10 +2,11 @@ package zio.logging
 
 import java.nio.charset.Charset
 import java.nio.file.Path
+import java.util.concurrent.atomic.AtomicBoolean
 
 import izumi.reflect.Tag
 import zio.console._
-import zio.{ Has, UIO, ULayer, URIO, ZIO, ZLayer, ZManaged, ZQueue }
+import zio.{ Has, Task, UIO, ULayer, URIO, ZIO, ZLayer, ZManaged, ZQueue }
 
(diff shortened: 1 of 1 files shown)


core/src/main/scala/zio/logging/LogAppender.scala
```diff
@@ -83,9 +84,16 @@ object LogAppender {
       .fromAutoCloseable(UIO(new LogWriter(destination, charset, autoFlushBatchSize, bufferedIOSize)))
       .map(writer =>
         new Service[A] {
+          private val hasWarned = new AtomicBoolean()
```
pshemass, line 87: can we use `Ref` instead?

pshemass (approved):

core/src/main/scala/zio/logging/LogAppender.scala
reibitto: Sure thing.

pshemass (approved):

PR #202, opened 2021-01-02
Slf4j bridge

vigoo:
This PR adds a new module called `zio-logging-slf4j-bridge`, which is the "opposite" of the existing SLF4j support, registering `zio-logging` as an SLF4j provider. 
With this third party libraries logging through SLF4j can log through zio-logging without the need to a third party slf4j implementation.

This is good for reducing dependencies and simplify the whole logging stack of services.
For example we are using this for mixing in sttp and Netty logs to the zio-logging log stream of a service.

Easy filtering configuration makes this much more usable, a proposed solution for that is coming in another PR.

### DIFF
--- a/slf4j-bridge/src/main/scala/org/slf4j/impl/ZioLoggerFactory.scala
+++ b/slf4j-bridge/src/main/scala/org/slf4j/impl/ZioLoggerFactory.scala
@@ -1,0 +1,30 @@
+package org.slf4j.impl
+
+import org.slf4j.{ ILoggerFactory, Logger }
+import zio.ZIO
+import zio.logging.Logging
+
+import java.util.concurrent.ConcurrentHashMap
+import scala.jdk.CollectionConverters._
+
+class ZioLoggerFactory extends ILoggerFactory {
+  private var runtime: zio.Runtime[Logging] = null
+  private val loggers = new ConcurrentHashMap[String, Logger]().asScala
+
+  def attachRuntime(runtime: zio.Runtime[Logging]): Unit =
+    this.runtime = runtime
+
+  private[impl] def run[A](f: ZIO[Logging, Nothing, A]): Unit =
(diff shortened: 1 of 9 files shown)


pshemass:
Thanks @vigoo for this PR it looks pretty good.

pshemass (approved):

slf4j-bridge/src/main/scala/org/slf4j/impl/ZioLoggerFactory.scala
```diff
@@ -0,0 +1,32 @@
...
+import scala.jdk.CollectionConverters._
+
+class ZioLoggerFactory extends ILoggerFactory {
+  private var runtime: zio.Runtime[Logging] = null
+  private val loggers                       = new ConcurrentHashMap[String, Logger]().asScala
+
+  def attachRuntime(runtime: zio.Runtime[Logging]): Unit =
+    this.runtime = runtime
+
+  private[impl] def run[A](f: ZIO[Logging, Nothing, A]): Unit =
```
jdegoes, line 17: Delete the `A` type parameter and accept `ZIO[Logging, Nothing, Any]`, instead. Clearer expression of the same idea.

slf4j-bridge/src/main/scala/org/slf4j/impl/ZioLoggerFactory.scala
```diff
@@ -0,0 +1,32 @@
...
+      runtime.unsafeRun(f)
+      ()
+    }
+
+  override def getLogger(name: String): Logger =
+    loggers.getOrElseUpdate(name, new ZioLogger(name, this))
+}
+
+object ZioLoggerFactory {
+  def bind(runtime: zio.Runtime[Logging]): Unit =
```
jdegoes, line 28: `initialize`?

PR #203, opened 2021-01-02
Making withLoggerNameFromLine public in core

vigoo:
I ran into wanting to reuse `withLoggerNameFromLine` when defining a custom appender, and had to copy it to my project as it's a private function in the `slf4j` module. 

This PR moves it to `coreJVM` and makes it public.


pshemass:
Hi @vigoo, I had this idea in past but @jdegoes didn't like it but it's new year maybe @jdegoes changed his mind :P

However I would prefer to not mix this in to `LogAppender` companion.

vigoo:
I mixed it there because it's a bit like `async` in a way that modifies an appender, and it's already there. But of course I'm happy to move it elsewhere. And of course happy to discuss with @jdegoes why is it bad to make it public; I agree that it's better to have explicit named log contexts and I guess the reason is to not encourage using this tracing based fallback. 

On the other hand the need to have a mandatory name is not necessarily slf4j specific, so it makes sense to move it to the core library. 

Now I tried to recall why this came up for me and realized that it was the following sequence of events:

- I started using zio-logging with slf4j logger
- Because of this fallback mechanism I was lazy and did not set up proper names in the log contexts, as it "just worked"
- Then I moved to pure zio-logging console appender and suddenly the names were missing
- During this step I wanted to just have the same output as before with slf4j+log4j2 so I copied the private function to have it
- Then later added the proper annotations everywhere and the copied function remained in the code as a "TODO: can it be public in the lib?" 

:) 

So I think to be consistent it would make sense to either have this building block public (as in this PR) _or_ not have it at all, not even in the slf4j logger.

jdegoes (approved):

jdegoes:
Good to merge when lint passes!

PR #204, opened 2021-01-02
A simple logger tree-based filtering

vigoo:
Introduces a simple hierarchic name-based filtering method that is similar to other logging systems. The filter is defined by a list of names tied to minimum log levels.


vigoo:
Very useful when using the SLF4j bridge (https://github.com/zio/zio-logging/pull/202)

core/src/main/scala/zio/logging/LogFiltering.scala
```diff
@@ -0,0 +1,83 @@
...
+   * @param mappings List of mappings, nesting defined by dot-separated strings
+   * @return A filter function for customizing appenders
+   */
+  def filterBy[A](rootLevel: LogLevel, mappings: (String, LogLevel)*): (LogContext, => A) => Boolean =
+    filterByTree(buildLogFilterTree(rootLevel, mappings))
+
+  def filterByTree[A](root: LogFilterNode): (LogContext, => A) => Boolean =
+    (ctx, _) => {
+      val loggerName = ctx.get(LogAnnotation.Name).flatMap(_.split('.'))
+      val logLevel   = findMostSpecificLogLevel(loggerName, root)
```
pshemass, line 35: did you check what will be performance impact of this?

core/src/main/scala/zio/logging/LogFiltering.scala
```diff
@@ -0,0 +1,83 @@
...
+      ctx.get(LogAnnotation.Level) >= logLevel
+    }
+
+  private def buildLogFilterTree(rootLevel: LogLevel, mappings: Seq[(String, LogLevel)]): LogFilterNode = {
+    def add(tree: LogFilterNode, names: List[String], level: LogLevel): LogFilterNode =
```
pshemass, line 40: we probably should use `Chunk` instead.

core/src/main/scala/zio/logging/LogFiltering.scala
vigoo: Not yet, I can add some benchmarks to have an understanding. And then it could be probably improved with a cache or something.

core/src/main/scala/zio/logging/LogFiltering.scala
vigoo: Yeah, but note that the `LogAnnotation.Name` has a type of `List[String]` too

pshemass: yeah, I forgot about. we should probably change that in the future

vigoo:
I created some benchmarks and also wrote a simple `TMap` based caching version around the filter function. The results:

```
   * FilterBenchmarks.handWrittenFilterEval     thrpt    5  9177150.646 ± 125715.644  ops/s
   * FilterBenchmarks.filterTreeEval            thrpt    5  7298406.870 ±  87773.959  ops/s

   * FilterBenchmarks.noFilteringLogging        thrpt    5   267066.692 ±   2170.339  ops/s
   * FilterBenchmarks.handWrittenFilterLogging  thrpt    5   262466.006 ±   3641.051  ops/s
   * FilterBenchmarks.filterTreeLog             thrpt    5   252841.756 ±   2912.062  ops/s
   * FilterBenchmarks.cachedFilterTreeLog       thrpt    5   260752.769 ±   2625.707  ops/s
```

For me this looks like the tree-based filter is comparable to a hand-written one (for a few nodes in the tree, which is the normal use case to my experience - thinking of the number of differently configured loggers in a typical log4j config for example).
The caching makes it almost as fast as the hand-written version, but in the benchmarks measuring logging not just the filter function itself it seems like the filtering itself is not that significant part of the whole operation. For this reason I did not completely finish the cached filtering version, meaning it is not that convenient yet to initialize it (requires the `TMap` from outside). If you think it worth it I can make the cache construction part of the appender layer etc.

core/src/main/scala/zio/logging/LogFiltering.scala
```diff
@@ -0,0 +1,114 @@
+package zio.logging
+
+import zio.stm.{ TMap, ZSTM }
+import zio.UIO
+
+import scala.annotation.tailrec
+
+object LogFiltering {
+  case class LogFilterNode(logLevel: LogLevel, children: Map[String, LogFilterNode])
```
jdegoes, line 9: `final`

vigoo:
@jdegoes I believe this is ready to merge

jdegoes (approved):

PR #205, opened 2021-01-02
Capturing a Cause->Throwable converter in the log annotation

vigoo:
This is a bit bigger change as it breaks the API but I believe something like this would make the library much easier to use with custom (non `Throwable`) error types. 

I bumped into this when tried to log errors through the SLF4j logger in a way they appear errors (exceptions) for the logger while still using more well typed errors on the ZIO level. The two primary reason why this is a good idea, I think is:
- The current SLF4j logger does nothing with the `Cause` annotation. This is what I first ran into, logging errors with `error(msg, cause)` and not seeing the cause in any way in the `zio-logging -> slf4j -> log4j2` chain's output. Without my change in the `slf4j` module the only thing you can do is `prettyPrint` the cause to a string, but then there is no straightforward way to pass it down to the underlying logger (except appending to the given message but that's weird)
- With some appenders it is more important to pass error causes in the "standard" way, as throwables attached to the log entries. For an example consider the Sentry appender.

So the change in this PR introduces an implicit conversion from `E` to `Throwable` for logging purposes, with default implementation for `E <: Throwable` and `Nothing`.

### DIFF
--- a/core/src/main/scala/zio/logging/CapturedCause.scala
+++ b/core/src/main/scala/zio/logging/CapturedCause.scala
@@ -1,0 +1,28 @@
+package zio.logging
+
+import zio.Cause
+
+sealed trait CapturedCause {
+  val cause: Cause[Any]
+  def toThrowable: Throwable
+}
+
+object CapturedCause {
+  def apply[E: CauseToThrowable](c: Cause[E]): CapturedCause =
+    new CapturedCause {
+      override val cause: Cause[Any]      = c
+      override def toThrowable: Throwable =
+        implicitly[CauseToThrowable[E]].toThrowable(c)
+    }
+
+  trait CauseToThrowable[-E] {
+    def toThrowable(cause: Cause[E]): Throwable
+  }
+
+  implicit def throwableCause[E <: Throwable]: CauseToThrowable[E] = (cause: Cause[E]) => cause.squash
+  implicit val nothingCause: CauseToThrowable[Nothing]             = (_: Cause[Nothing]) => new RuntimeException("Nothing")
+
+  object causeToRuntimeException {
+    implicit def toThrowable[E]: CauseToThrowable[E] = (cause: Cause[E]) => new RuntimeException(cause.prettyPrint)
+  }
+}
--- a/core/src/main/scala/zio/logging/Logger.scala
+++ b/core/src/main/scala/zio/logging/Logger.scala
@@ -1,5 +1,6 @@
 package zio.logging
 
+import zio.logging.CapturedCause.CauseToThrowable
 import zio.{ Cause, FiberRef, UIO, URIO, ZIO }
 
 trait Logger[-A] { self =>
@@ -43,15 +44,16 @@ trait Logger[-A] { self =>
   /**
    * Logs the specified element at the error level with cause.
    */
-  def error(line: => A, cause: Cause[Any]): UIO[Unit] =
-    self.locally(LogAnnotation.Cause(Some(cause))) {
+  def error[E: CauseToThrowable](line: => A, cause: Cause[E]): UIO[Unit] =
+    self.locally(LogAnnotation.Cause(Some(CapturedCause(cause)))) {
       self.log(LogLevel.Error)(line)
     }
 
   /**
    * Evaluates the specified element based on the LogLevel set and logs at the error level
    */
-  def errorM[R, E](line: ZIO[R, E, A], cause: Cause[Any]): ZIO[R, E, Unit] = line.flatMap(l => error(l, cause))
+  def errorM[R, E: CauseToThrowable](line: ZIO[R, E, A], cause: Cause[E]): ZIO[R, E, Unit] =
+    line.flatMap(l => error(l, cause))
 
   /**
    * Derives a new logger from this one, by applying the specified decorator
--- a/core/src/main/scala/zio/logging/LogAnnotation.scala
+++ b/core/src/main/scala/zio/logging/LogAnnotation.scala
@@ -1,9 +1,6 @@
 package zio.logging
 
 import java.time.OffsetDateTime
-
-import zio.Cause
-
 import scala.reflect.ClassTag
 
 /**
@@ -70,11 +67,11 @@ object LogAnnotation {
    * The `Cause` annotation keeps track of a Cause.
    */
   val Cause =
-    LogAnnotation[Option[Cause[Any]]](
+    LogAnnotation[Option[CapturedCause]](
       name = "cause",
       initialValue = None,
       combine = (_, t) => t,
-      _.map(_.prettyPrint).getOrElse("")
+      _.map(_.cause.prettyPrint).getOrElse("")
     )
 
   /**
--- a/core/src/main/scala/zio/logging/LogFormat.scala
+++ b/core/src/main/scala/zio/logging/LogFormat.scala
@@ -33,9 +33,9 @@ object LogFormat {
       val loggerName = context(LogAnnotation.Name)
       val maybeError = context
         .get(LogAnnotation.Throwable)
-        .map(Cause.fail)
+        .map(throwable => CapturedCause(Cause.fail(throwable)))
         .orElse(context.get(LogAnnotation.Cause))
-        .map(cause => NL + cause.prettyPrint)
+        .map(cause => NL + cause.cause.prettyPrint)
         .getOrElse("")
       date + " " + level + " " + loggerName + " " + format0(context, line) + " " + maybeError
     }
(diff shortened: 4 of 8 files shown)


core/src/main/scala/zio/logging/CapturedCause.scala
```diff
@@ -0,0 +1,28 @@
...
+
+  trait CauseToThrowable[-E] {
+    def toThrowable(cause: Cause[E]): Throwable
+  }
+
+  implicit def throwableCause[E <: Throwable]: CauseToThrowable[E] = (cause: Cause[E]) => cause.squash
+  implicit val nothingCause: CauseToThrowable[Nothing]             = (_: Cause[Nothing]) => new RuntimeException("Nothing")
+
+  object causeToRuntimeException {
+    implicit def toThrowable[E]: CauseToThrowable[E] = (cause: Cause[E]) => new RuntimeException(cause.prettyPrint)
```
pshemass, line 26: I would probably add this to `Format` handle this. I don't like to make this decision on compile time. Imagine situation that you create library and you decided in your code to use `causeToRuntimeException` and other user try to use your library be it has different idea. If that will be decision that you could make in format it's more flexible.

pshemass (changes requested):

core/src/main/scala/zio/logging/Logger.scala
```diff
@@ -43,15 +44,16 @@ trait Logger[-A] { self =>
   /**
    * Logs the specified element at the error level with cause.
    */
-  def error(line: => A, cause: Cause[Any]): UIO[Unit] =
-    self.locally(LogAnnotation.Cause(Some(cause))) {
+  def error[E: CauseToThrowable](line: => A, cause: Cause[E]): UIO[Unit] =
```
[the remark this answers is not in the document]
vigoo: My intention was to be able to write well typed `E => Throwable` transformations because `E` gets erased when it's put into the annotation. So this is the last place where it can be captured.

core/src/main/scala/zio/logging/CapturedCause.scala
vigoo: This is a good point, I was not really thinking over how this would be used when logging from a library. Yeah if the library has a custom error type and its author used that fallback instead of providing a proper implementation, then there is no way to override it from the application. On the other hand if it is handled in the formatter, that would require to handle all the possible custom error types in a common log formatter defined in the application, even if it those are implementation detail of the given library that logs it. Maybe this would not be a very common situation, but I still feel that my proposed solution may work better.

core/src/main/scala/zio/logging/Logger.scala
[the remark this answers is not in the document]
pshemass: we can change annotation or create new one for this purpose.

core/src/main/scala/zio/logging/CapturedCause.scala
pshemass: it will required the same amount of work to do this in format so please add annotation that will carry error and handle this in format. @jdegoes if you have any idea how to make nice api for that please weigh in.

vigoo: I would like to summarize one more time why I believe that my proposed solution to do this on the logger side is better.

I won't talk about _why_ is it good to be able to have a conversion to Throwable because I think we are on the same page about that. If not I'm happy to write in more details about that one too.

Let's imagine a situation of writing a service that uses a (imagined version of) `zio-web` to implement a HTTP API with some library-provided, zio-logging based logging middleware. It also uses `zio-aws` to work with some AWS services such as S3 and also have its own debug logging for that. It can also use an internal company specific library to work with some database let's say, with company standard logging included, and the service itself has it's own business logic. In this example I would imagine the HTTP library has an error type that is somehow the union of some HTTP errors and the middleware errors, the AWS library has it's own error type, let's say the SQL library works with Throwables as it uses JDBC underneath, and the service business logic also has it's own typed errors. 

In this example each component would use zio-logging independently to log errors of different types. 

With my proposed solution each library's developer can decide to use the deault conversion or provide a more specific one. For example one of the error cases wraps exceptions coming from the AWS Java library, so that can be converted in a better way to throwable, keeping the nested exception. This is something that should be decided by the developer of the library, and written once. With that version the service developer just pulls in the various libraries using logging, provide a logging layer and everything just works automatically.

On the other hand, if I understand correctly, you propose to handle this conversion in the log formatter. I see multiple problems with that.

First of all, as I said, the error type gets erased currently. I see three ways to fix that. Either capture the conversion itself at the entry point (this is my proposal), or just capture a type tag, or convert the annotations to a HList or similar. 

Let's try to imagine what it would look like if we choose one of the other possibilities.

If we just store a `Cause[_]` and a type tag representing `E`, in your formatter you have to check the type tag to "reverse engineer" the original base error type, cast the cause, and then do the desired conversion. And every time you add anything new to the service, include a new library, or someone writes a new functionality, this has to be reviewed because there is no way to compile time check that all the possible `E`s are covered. And you need to replicate the knowledge of what a 3rd party library is logging in each application that uses that library.

With a `HList`-like structure you could have a more or less convenient way to choose is to select an implicit conversion matching the `E` of the annotation's `Cause` on the formatter side. But to do that, you would have to parametrize on the HList type all the way down to the formatter, which is much more invasive than capturing at the log entrypoint. And even if it would be done, it has the same problem I mentioned earlier of having a central place to know about all the error types in the application. (Although with compile time type checks, with most likely very hard to understand errors for incomplete definitions)

So I really don't see how that could work. If you have any idea that I missed, please let me know.

jdegoes:
@vigoo Can we simply always wrap the `Cause` in `FiberFailure`? That's the idiomatic way to go from `Cause` to `Throwable`.

core/src/main/scala/zio/logging/Logger.scala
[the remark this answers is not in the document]
jdegoes: If this is required, and I am not sure that it is (see question on the PR), then the idiomatic solution is to require `Cause[Throwable]` here, rather than to accept `Cause[E]` and a function `E => Throwable`.

If the user can go from `E` to `Throwable`, then they can just `map` over the `Cause[E]` to turn it to `Cause[Throwable]` before passing it into this function.

core/src/main/scala/zio/logging/Logger.scala
```diff
@@ -43,15 +44,16 @@ trait Logger[-A] { self =>
...
+    self.locally(LogAnnotation.Cause(Some(CapturedCause(cause)))) {
       self.log(LogLevel.Error)(line)
     }
 
   /**
    * Evaluates the specified element based on the LogLevel set and logs at the error level
    */
-  def errorM[R, E](line: ZIO[R, E, A], cause: Cause[Any]): ZIO[R, E, Unit] = line.flatMap(l => error(l, cause))
+  def errorM[R, E: CauseToThrowable](line: ZIO[R, E, A], cause: Cause[E]): ZIO[R, E, Unit] =
```
jdegoes, line 55: Same for this one.

core/src/main/scala/zio/logging/CapturedCause.scala
```diff
@@ -0,0 +1,28 @@
...
+
+object CapturedCause {
+  def apply[E: CauseToThrowable](c: Cause[E]): CapturedCause =
+    new CapturedCause {
+      override val cause: Cause[Any]      = c
+      override def toThrowable: Throwable =
+        implicitly[CauseToThrowable[E]].toThrowable(c)
+    }
+
+  trait CauseToThrowable[-E] {
```
jdegoes, line 18: I think we should delete this for reasons covered below.

core/src/main/scala/zio/logging/CapturedCause.scala
```diff
@@ -0,0 +1,28 @@
+package zio.logging
+
+import zio.Cause
+
+sealed trait CapturedCause {
```
jdegoes, line 5: I think we can delete this for reasons covered below.

vigoo:
I think I missed `FiberFalure`. Seems like a good default behavior, so the default implementation that I called `throwableCause` should definitely use that instead.

Modifying the slf4j adapter to log cause annotations with `FiberFailure` would solve the original issue more or less.

Additional to that I still believe that ability to define custom conversions for given error types is a plus, when using zio-logging together with more special log appenders through the slf4j interface. The best example I'm aware of is the [Sentry](https://sentry.io/welcome/) appender but I believe there are other ones out there too. 
These use the exception passed to the log appender specially and aggregate issues by it, show it on a UI etc. For those, it makes sense to let's say have an

```
E = DomainFailure(ErrorADT) | CommunicationFailure(IOException)
```

and be able to convert the `CommunicationFailure` to a nested exception passed to the logger, to preserve all the information.

So I wanted to solve two things with the PR:
- the surprising behavior that I log a cause and use an slf4j logger and I can't see the cause by default on the output; for that automatically converting to FiberFailure is a good solution
- being able to preserve captured exceptions through custom error types

vigoo:
After discussing this with @jdegoes we decided that a better approach would be to encode the error type in `Logger` and require it to be `Throwable` for SLF4j interop. I'm closing this PR and will open a new one exploring that version.

PR #206, opened 2021-01-02
AssembledLogFormat for building log formats with a small DSL

vigoo:
This is a small optional DSL to build `LogFormat`s


pshemass:
> This is a small optional DSL to build `LogFormat`s

this looks promising but we will need couple of examples in docs.

core/src/main/scala/zio/logging/LogFormat.scala
```diff
@@ -82,4 +84,80 @@ object LogFormat {
       format(lineFormat(context, line), date, level, loggerName, maybeError)
     }
   }
+
+  object AssembledLogFormat {
+    type FormatterFunction = (StringBuilder, LogContext, String) => Any
```
jdegoes, line 89: This should be a case class so you don't have to add syntax. In addition, the case class should store a function, but probably that function should be package private.

vigoo: Sounds good, I will refactor

core/src/main/scala/zio/logging/LogFormat.scala
```diff
@@ -82,4 +84,80 @@ object LogFormat {
...
+        (builder, ctx, _) => builder.append(f(ctx(annotation)))
+
+      def annotationF[A](annotation: LogAnnotation[A], f: A => String): FormatterFunction = (builder, ctx, _) =>
+        builder.append(f(ctx.get(annotation)))
+
+      def bracketed(inner: FormatterFunction): FormatterFunction =
+        bracketStart >>> inner >>> bracketEnd
+
+      implicit class FormatterFunctionOps(f: FormatterFunction) {
+        def >>>(g: FormatterFunction): FormatterFunction =
```
jdegoes, line 112: I think the best operator to use here is `+`, similar to string concatenation.

core/src/main/scala/zio/logging/LogFormat.scala
```diff
@@ -82,4 +84,80 @@ object LogFormat {
...
+          (builder, ctx, line) => {
+            f(builder, ctx, line)
+            g(builder, ctx, line)
+          }
+
+        def >+>(g: FormatterFunction): FormatterFunction =
```
jdegoes, line 118: I'd probably do `l.spaced(r)` for this one (i.e. no operator).

core/src/main/scala/zio/logging/LogFormat.scala
```diff
@@ -82,4 +84,80 @@ object LogFormat {
...
+          (builder, ctx, line) => {
+            f(builder, ctx, line)
+            builder.append(' ')
+            g(builder, ctx, line)
+          }
+      }
+
+      def level: FormatterFunction =
```
jdegoes, line 126: Should be `val` and not `def` to avoid constant re-creation.

vigoo:
@jdegoes I made the proposed changes with the following notes:
- I kept (but renamed) the operator version of `spaced` while also providing a named method. I think it is nice to have it if we also have `+` for concatenation. At least for me, `a <+> b + c` seems to be a readable and understandable description of spaced vs non-spaced concatenation. Of course this is very subjective, so if you don't agree I'll delete it.
- @pshemass I am happy to write docs about this later but I think the current "single-page" microsite is already too long for that, would be nicer to have a multi-page site where this can be added as a separate section. Would it be ok to do that separately? I'm happy to do it/help with it when I have time.

jdegoes:
The operator `<+>` means applicative plus in some contexts, and the angular brackets are almost always used for applicative / monad operators. I'm not sure if that could be confusing or not.

jdegoes (approved):

core/shared/src/main/scala/zio/logging/LogFormat.scala
```diff
@@ -82,4 +84,100 @@ object LogFormat {
       format(lineFormat(context, line), date, level, loggerName, maybeError)
     }
   }
+
+  object AssembledLogFormat {
+    case class FormatterFunction(private[logging] val fun: (StringBuilder, LogContext, String) => Any) {
```
jdegoes, line 89: Can we make this `final`?

pshemass:
> @jdegoes I made the proposed changes with the following notes:
> 
> * I kept (but renamed) the operator version of `spaced` while also providing a named method. I think it is nice to have it if we also have `+` for concatenation. At least for me, `a <+> b + c` seems to be a readable and understandable description of spaced vs non-spaced concatenation. Of course this is very subjective, so if you don't agree I'll delete it.
> * @pshemass I am happy to write docs about this later but I think the current "single-page" microsite is already too long for that, would be nicer to have a multi-page site where this can be added as a separate section. Would it be ok to do that separately? I'm happy to do it/help with it when I have time.

@vigoo I agree this doc is screaming for change :) If you promise that you gonna add it later I'm fine with this.

pshemass (approved):

vigoo:
> 
> 
> The operator `<+>` means applicative plus in some contexts, and the angular brackets are almost always used for applicative / monad operators. I'm not sure if that could be confusing or not.

I'm not sure either. Just realized that I think for me it seems natural because of using [this pretty printer library a lot](https://hackage.haskell.org/package/wl-pprint) in the past. Although it uses the `<>` for the concatenation (just as the original Wadler paper) so I guess this one is confusing even for people used to that :)

On the other hand I find this nicer:
```scala
bracketed(LEVEL) <+> timestamp(formatter) <+> name <+> line + error
```

than this:

```scala
bracketed(LEVEL) spaced timestamp(formatter) spaced name spaced line + error
```

pshemass:
> > The operator `<+>` means applicative plus in some contexts, and the angular brackets are almost always used for applicative / monad operators. I'm not sure if that could be confusing or not.
> 
> I'm not sure either. Just realized that I think for me it seems natural because of using [this pretty printer library a lot](https://hackage.haskell.org/package/wl-pprint-1.2.1/docs/Text-PrettyPrint-Leijen.html#v:-60--43--62-) in the past. Although it uses the `<>` for the concatenation (just as the original Wadler paper) so I guess this one is confusing even for people used to that :)
> 
> On the other hand I find this nicer:
> 
> ```scala
> bracketed(LEVEL) <+> timestamp(formatter) <+> name <+> line + error
> ```
> 
> than this:
> 
> ```scala
> bracketed(LEVEL) spaced timestamp(formatter) spaced name spaced line + error
> ```

@jdegoes @vigoo what do think about this

`bracketed(LEVEL) |_| timestamp(formatter) |_| name |_| line + error`

pshemass:
@vigoo any update on that?

vigoo:
@pshemass @jdegoes I have no strong feelings about these operators :) so if we agree in any I'm happy to change. Personally I like my original (that's currently in the PR) the best so please decide.

jdegoes:
I like `|_|`!

vigoo:
The `|_|` is invalid but I changed it to `|-|`

jdegoes (approved):

PR #233, opened 2021-02-28
Ensure log message laziness in Slf4jLogger with logging guards

aartigao:
Added logging guards to prevent log message evaluation if current logger is disabled.

Currently if you have a potentially expensive interpolated `String` like:

```scala
log.debug(s"Pushing: $datums")
```

where `$datums`'s `toString` is an expensive operation it gets evaluated even if the log level is disabled. This evaluation takes place at the boundary between Scala (ZIO Logging) and Java (SLF4J Logging).

This PR prevents that by setting logging guards before SLF4J Logging invocation, making the operations in ZIO side truly lazy.


pshemass (approved):

pshemass:
@aartigao thanks for that fix

Issue #240, opened 2021-03-19
FR: slf4j bridge improvement


virusdave:
Slf4j docs show a pattern of using `{}` in log format message (see http://www.slf4j.org/manual.html#typical_usage)
However, the bridge implementation uses `String.format` everywhere (see eg. https://github.com/zio/zio-logging/blob/master/slf4j-bridge/src/main/scala-2/org/slf4j/impl/ZioLogger.scala#L20), which doesn't substitute the arguments' `.toString` result in place of that when logging.  The result for many cases is logs filled with stuff like `org.eclipse.jetty.util.component.AbstractLifeCycle] STARTED @{}ms {}`, which is less than helpful :)

Please support positional arguments?  Perhaps just replacing `{}` by `%s` in the format string would be sufficient?

jdegoes:
The SLF4J bridge needs to be created in the 2.0 branch. Hopefully this ticket can incorporated into the redesign.

justcoon:
fixed in #482

PR #241, opened 2021-03-19
Handle positional arguments in slf4j bridge

virusdave:
See title.
Handles issue https://github.com/zio/zio-logging/issues/240

### DIFF
--- a/slf4j-bridge/src/main/scala-2/org/slf4j/impl/ZioLogger.scala
+++ b/slf4j-bridge/src/main/scala-2/org/slf4j/impl/ZioLogger.scala
@@ -1,6 +1,6 @@
 package org.slf4j.impl
 
-import org.slf4j.helpers.MarkerIgnoringBase
+import org.slf4j.helpers.{ MarkerIgnoringBase, MessageFormatter }
 import zio.ZIO
 import zio.logging.{ log, LogAnnotation, Logging }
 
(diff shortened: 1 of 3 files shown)


pshemass:
@vigoo could you take a look on that ?

vigoo:
Seems like `slf4j-api` has its own `MessageFormatter` that implements slf4j's logic and is faster than the standard string formatter (especially with the extra replace):
https://www.javadoc.io/doc/org.slf4j/slf4j-api/1.7.30/org/slf4j/helpers/MessageFormatter.html

I think we should use this one. @virusdave can you try it?

virusdave:
> Seems like `slf4j-api` has its own `MessageFormatter` that implements slf4j's logic and is faster than the standard string formatter (especially with the extra replace):
> https://www.javadoc.io/doc/org.slf4j/slf4j-api/1.7.30/org/slf4j/helpers/MessageFormatter.html
> 
> I think we should use this one. @virusdave can you try it?

Moving this week, so not sure when I'll have time to get to it, but it looks simple enough to do (`s/String.format/MessageFormatter.format` in the original source).  If no one gets to it first, I'll do it, but it'll take me a couple days.

pshemass:
@virusdave any update on that?

aartigao:
This is a nice feature! ❤️ 

Any plan on fixing the conflicts and merging it @virusdave?

jdegoes:
Will close for now, feel free to re-open against 2.0 work!

PR #289, opened 2021-06-15
Add LogAnnotation.optional

iravid:
The majority of the annotations I use are of the form `LogAnnotation[Option[A]]`; writing out the last value combiner and getOrElse("") as the renderer is pretty repetitive so this constructor will save a bunch of characters.


pshemass (approved):

pshemass:
@iravid thanks for PR. I will release this right away.

iravid:
@pshemass thanks! 🙏🏻

PR #336, opened 2021-10-04
Add getAll function to LogContext

duxet:
This should make it easier to integrate LogContext with some other stuff like error reporting or tracing

### DIFF
--- a/core/shared/src/main/scala/zio/logging/LogContext.scala
+++ b/core/shared/src/main/scala/zio/logging/LogContext.scala
@@ -33,6 +33,12 @@ final case class LogContext private (private val map: Map[LogAnnotation[_], Any]
     map.get(annotation).fold(annotation.initialValue)(_.asInstanceOf[A])
 
   /**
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/zio/logging/LogContext.scala
```diff
@@ -32,6 +32,12 @@ final case class LogContext private (private val map: Map[LogAnnotation[_], Any]
   def get[A](annotation: LogAnnotation[A]): A =
     map.get(annotation).fold(annotation.initialValue)(_.asInstanceOf[A])
 
+  /**
+   * Retrieves all annotations from the context.
+   */
+  def getAll: Map[LogAnnotation[_], Any] =
```
pshemass, line 38: returning something with `Any` make me feel sick :) 

@jdegoes what do you think?

jdegoes:
This is a very useful feature, which should be easy in the new design. Will close for now but thanks for your work identifying and fixing this issue!

PR #363, opened 2021-10-31
DSL for format and slf4j backend

pshemass:


backend/jvm/src/main/scala/zio/logging/backend/NewZIOLogging.scala
```diff
@@ -0,0 +1,19 @@
+package zio.logging.backend
```
jdegoes, line 1: Probably we need an "examples" project that contains this file.

backend/jvm/src/main/scala/zio/logging/backend/PlatformSpecificBackends.scala
```diff
@@ -0,0 +1,69 @@
...
+
+import scala.jdk.CollectionConverters._
+
+trait PlatformSpecificBackends {
+
+  def slf4j(
+    logLevel: zio.LogLevel,
+    format: LogFormat[String],
+    rootLoggerName: ZTraceElement => String
+  ): RuntimeConfigAspect =
```
jdegoes, line 15: So nice!

backend/jvm/src/main/scala/zio/logging/backend/PlatformSpecificBackends.scala
```diff
@@ -0,0 +1,69 @@
...
+        spans: List[LogSpan]
+      ): Unit =
+        formatLogger(trace, fiberId, logLevel, message, context, spans).foreach { message =>
+          val slf4jLogger = LoggerFactory.getLogger(rootLoggerName(trace))
+
+          val previous =
+            context.get(logAnnotation) match {
+              case Some(annotations: Map[String, String] @unchecked) if annotations.nonEmpty =>
+                MDC.setContextMap(annotations.asJava)
+                Some(Option(MDC.getCopyOfContextMap).getOrElse(java.util.Collections.emptyMap[String, String]()))
```
jdegoes, line 52: This is copying the context map AFTER it has updated it to the log annotations. If you are trying to restore it to the old value, then I would do that prior to setting the context map.

backend/shared/src/main/scala/zio/logging/LogFormat.scala
```diff
@@ -0,0 +1,214 @@
...
+    override private[logging] def unsafeFormat(builder: Out2 => Unit) =
+      self.unsafeFormat((out: Out) => builder(fn(out)))
+
+    override def toLogger: ZLogger[Out2] = self.toLogger.map(fn)
+  }
+
+  final def spaced(other: LogFormat[String])(implicit ev: Out <:< String): LogFormat[String] =
+    this |-| other
+
+  def toLogger: ZLogger[Out]
```
jdegoes, line 87: Rewrite this as follows:

```scala
def toLogger[A1 >: A](lft: LogFormatType[A1]): ZLogger[A1]
```

where:

```scala
final case class LogFormatType[A](unsafeMake: LogFormatBuilder[A, A])
object LogFormatType {
  val string: LogFormatType[String] = ???
}

final case class LogFormatBuilder[-In, +Out](unsafeAppend: In => Unit, unsafeResult: () => Out)
```

Because you need 3 things to implement `toLogger` generically:

1. Make a new builder for that type.
2. Given a builder:
   1. Append something to it
   2. Build the final result

The above case classes will give you what you need to do the generic `toLogger`.

backend/shared/src/main/scala/zio/logging/LogFormat.scala
```diff
@@ -0,0 +1,214 @@
...
+    bracketStart + inner + bracketEnd
+
+  val level: LogFormat[String] =
+    LogFormat { (builder, _, _, level, _, _, _) =>
+      builder(level.label)
+    }
+
+  val fiberNumber: LogFormat[String] =
+    LogFormat { (builder, _, fiberId, _, _, _, _) =>
+      builder("#")
```
jdegoes, line 160: We are standardizing on `zio-fiber-<seq-num>` to format the id of ZIO fibers.

backend/shared/src/main/scala/zio/logging/LogFormat.scala
```diff
@@ -0,0 +1,214 @@
...
+
+  val timestamp: LogFormat[String] = timestamp(DateTimeFormatter.ISO_OFFSET_DATE_TIME)
+
+  def timestamp(formatter: DateTimeFormatter): LogFormat[String] =
+    string {
+      val now = ZonedDateTime.now()
+      formatter.format(now)
+    }
+
+  val line: LogFormat[String] =
```
jdegoes, line 198: I would order all methods alphabetically, both in the companion object, and inside the `LogFormat` trait.

backend/shared/src/main/scala/zio/logging/LogFormat.scala
jdegoes, line 87: Rewrite this as follows:

```scala
def toLogger[A1 >: A](lft: LogFormatType[A1]): ZLogger[A1]
```

where:

```scala
final case class LogFormatType[A](unsafeMake: LogFormatBuilder[A, A])
object LogFormatType {
  val string: LogFormatType[String] = ???
}

final case class LogFormatBuilder[-In, +Out](unsafeAppend: In => Unit, unsafeResult: () => Out)
```

Because you need 3 things to implement `toLogger` generically:

1. Make a new builder for that type.
2. Given a builder:
   1. Append something to it
   2. Build the final result

The above case classes will give you what you need to do the generic `toLogger`.

backend/shared/src/main/scala/zio/logging/package.scala
```diff
@@ -0,0 +1,22 @@
+package zio
+
+package object logging {
```
jdegoes, line 3: Remember to PR to ZIO!

jdegoes (changes requested):
Looking great! Just a few minor comments.

backend/jvm/src/main/scala/zio/logging/backend/PlatformSpecificBackends.scala
pshemass: my bad! I still need to add test for this

jdegoes:
@pshemass Thinking about the complexity of the type parameter:

Maybe we should just specialize `LogFormat` to `String`.

Then we can add `StructuredLogFormat` which has the same API but is designed for JSON-like data, using the  `DynamicValue` type from ZIO Schema.

backend/shared/src/main/scala/zio/logging/LogFormat.scala
pshemass: why do you think we need `LogFormatType` ? at first glance it looks a bit redundant

backend/shared/src/main/scala/zio/logging/LogFormat.scala
pshemass: ack

backend/shared/src/main/scala/zio/logging/LogFormat.scala
jdegoes: You need it because you need 1 builder per each log message, so you need a way to make the builders.

Anyway, all of this is irrelevant: See my latest comment on the whole PR. I think we should make `LogFormat` monomorphic again, specialized to `String`, and introduce a corresponding `StructuredLogFormat` that is based on ZIO Schema's `DynamicValue` (or even just `Map[String, String]`).

jdegoes: This comment:

<https://github.com/zio/zio-logging/pull/363#issuecomment-956431679>

backend/shared/src/main/scala/zio/logging/package.scala
pshemass: I will merge this first and then open PRs for annotation and `Cause` in `Zlogger`

backend/shared/src/main/scala/zio/logging/LogFormat.scala
pshemass: unless you meant `final case class LogFormatType[A](unsafeMake: () => LogFormatBuilder[A, A])`

backend/shared/src/main/scala/zio/logging/LogFormat.scala
pshemass: `default` and `colored` need to be at the end because of initialization unless we are ok with `lazy val`

backend/jvm/src/main/scala/zio/logging/backend/NewZIOLogging.scala
pshemass: It will part of docs

pshemass:
> String

JSON is string at the end of the day so I think it should be fine. 

I would solve structural log with just different format something like 

```
json (
  field("@timestamp", timestamp),
  field("location", enclosingClass),
  annotations,
  ...
)
```

we may change `StringBuilder` with `Writer` to be able to use `zio-json`, what do you think about that?

backend/shared/src/main/scala/zio/logging/LogFormat.scala
pshemass: @jdegoes I agree with just different log format that will be for json and produce string.

backend/shared/src/main/scala/zio/logging/LogFormat.scala
```diff
@@ -0,0 +1,352 @@
...
+ * }}}
+ */
+trait LogFormat { self =>
+  private[logging] def unsafeFormat(
+    builder: String => Unit
+  ): ZLogger[Unit]
+
+  private[logging] def unsafeEncodeJson(
+    builder: String => Unit
+  ): ZLogger[Unit] = unsafeFormat(builder)
```
pshemass, line 24: @jdegoes do you think we need separate Json format or we can do it this way?

jdegoes (approved):

PR #393, opened 2022-01-06
Migrate to ZIO 2

adamgfraser:
Migrates ZIO Logging to ZIO 2.0. There are several backends that still need to be implemented but I think this gets the API to where we want to be.


pshemass:
LGTM. 

we were discussing with John moving `zio.logging.logAnnotation` FiberRef and `annotate` from https://github.com/zio/zio-logging/blob/zio2/backend/shared/src/main/scala/zio/logging/package.scala#L5 to Core  what do you think about this?

adamgfraser:
Yes, I actually had a PR that did that. See zio/zio#6148. I think @jdegoes had some concerns about the rendering of the annotations really being tied to the logging backend. I think the design space is also a little more open here so it could be valuable to have it in `zio-logging` where it could evolve independently. But can see advantages and disadvantages both ways.

jdegoes:
Perhaps a good middleground is adding a `logAnnotation` `FiberRef` as a `Map[String, String]`, then a `ZIO.logAnnotate` and associated annotation for string-keyed, string-valued annotations? That's probably 90% of log annotations anyway.

/cc @adamgfraser @pshemass

jdegoes:
Also: it's only 2 lines of code!

adamgfraser:
@jdegoes That sounds reasonable to me.

I think it ends up being a little more that that because I think in `ZLogger` we probably want to provide "first class" support for accessing the log annotations without having to go through the map of all `FiberRef` values.

The one disadvantage of this is it to some extent creates two ways to do log annotations, `ZIO.logAnnotate` and friends for "simple" annotations and `zio.logging.logAnnotation` for "fancy" annotations.

pshemass:
@jdegoes I totally agree that keeping annotation as `Map[String, String]` will be better. This idea that we implemented before was awkward in couple of ways like optional values for `correlationId` or we the idea to keep logger name as annotation (it was cool but not always practical).

adamgfraser:
Yeah I agree the treatment of optional values was awkward. I will update my PR accordingly.

PR #415, opened 2022-02-10
Readding slf4j-bridge

vigoo:
This PR readds the slf4j-bridge subproject, which enables users to use a lightweight, ZIO based logger (such as the console logger) without including any Java logging frameworks, but still capture the 3rd party library logs via SLF4j.


jdegoes:
Can you run Scalafix and Scalafmt?

jdegoes:
> We should have an official test logger backend that captures the log entries

👍 Yes. Possibly this should be bundled into ZIO Test so you can automatically access logs from your tests.

> It feels a bit unusual to not be able to log a message together with a throwable, although I worked around that by wrapping the failure in one more exception that encapsulates the custom message

See discussion on Slack.

> I'm not sure how setting up the whole thing will look like in the real world, so maybe having the bridge initialization as a ZIOApp is not a good idea. 

I think it should be a `RuntimeConfigAspect`, instead. What do you think about that?

jdegoes (approved):

PR #458, opened 2022-05-28
Json appender

davidlar:
This is a complete rework of https://github.com/zio/zio-logging/pull/424 after review by @jdegoes.

### DIFF
--- a/core/jvm/src/test/scala/zio/logging/JsonLogFormatSpec.scala
+++ b/core/jvm/src/test/scala/zio/logging/JsonLogFormatSpec.scala
@@ -1,0 +1,139 @@
+package zio.logging
+
+import zio.logging.LogFormat.{ line, _ }
+import zio.logging.internal.JsonEscape
+import zio.test._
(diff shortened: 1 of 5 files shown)


jdegoes (dismissed):
Looks phenomenal! Thanks for putting up with the revisions.

After conflicts are fixed, this is good to merge!

There is some separate work we could do later to optimize performance.

jdegoes:
@davidlar This is good to merge when you get a chance to fix conflicts!

davidlar:
Great!

I thought about reusing the string builder more instead of creating new ones. Don't know if it will affect performance or not.
What were the things you were thinking about?

jdegoes (approved):

jdegoes:
@davidlar Thanks for your persistence! We finally got this merged in. 

One of the main things, I think, would be using low-level 'while' loops and working directly with `Char`, because the current implementation will allocate and box a lot for escaping. In addition, we could go further with using CharSequence strategically.

PR #464, opened 2022-06-26
Update to ZIO 2.0.0

pawelkaczor:


core/jvm/src/test/scala/zio/logging/LogFormatSpec.scala
```diff
@@ -93,13 +92,16 @@ object LogFormatSpec extends ZIOSpecDefault {
     test("annotation (structured)") {
       val format = annotation(LogAnnotation.UserId)
       check(Gen.string) { annotationValue =>
-        val result = format.toLogger(
+        val fiberId: FiberId.Runtime = Unsafe.unsafeCompat(implicit u => FiberId.make(implicitly[Trace]))
```
adamgfraser, line 95: Don't think you should need the `implicitly[Trace]` here.

adamgfraser:
@pawelkaczor Looks like @landlockedsurfer beat you to it. If there is anything that you think is missing there feel free to submit a PR! Thank you for your work on this! 🙏

PR #465, opened 2022-06-30
Try to optimize json encoding a bit

davidlar:
@jdegoes Do you mean something like this?


jdegoes (approved):

jdegoes:
Looks good! We should also aim to reuse string builders if possible, and maybe have a benchmark that we can use to measure these sorts of optimizations. Thanks for your work on this!

PR #460, opened 2022-06-02
Fix file loggers

guersam:
Use `LogWriter#writeln` to flush lines properly

PR #466, opened 2022-07-04
Fix file loggers

guersam:
Supersedes #460


justcoon (approved):
@guersam thank you very much for contribution

Issue #453, opened 2022-05-12
ZIO.logAnnotate and LogAnnotation behave differently


Ellzord:
For the SLF4J logger + logstash the following code gives inconsistent results:

```scala
ZIO.logInfo("hello world") @@ ZIOAspect.annotated("username", "Ellzord") @@ LogAnnotation.UserId("1234")
```

```json
{
  "@timestamp": "2022-05-12T10:42:22.596+01:00",
  "@version": "1",
  "message": "hello world user_id=1234",
  "logger_name": "zio-slf4j-logger",
  "thread_name": "zio-default-blocking-2",
  "level": "INFO",
  "level_value": 20000,
  "username": "Ellzord"
}
```

Using `LogAnnotation` it'll only appear if you provide that to the `LogFormat` and then it'll be only in the `message`. While any annotations are given their own place in the JSON. I believe they should work the same and it should not be included in the message at all (i.e. no need to supply to `LogFormat` and they're counted as annotations). The naming is confusing if they are meant not to be the same thing.

It may also be possible to remove the need for the `LogContext` ref if they can become annotations. They would then share most of the same implementation and it doesn't matter if the user wants to use `ZIO.logAnnotate` or `LogAnnotation` (with a non-string type).

Currently I feel the only usable option is `ZIO.logAnnotate` as I don't believe the `message` should be mutated and it helps for indexing.

justcoon:
propagation of annotations to MDC context was fixed in `2.1.0` version, 
at this level zio LogAnnotation-s and zio-logging LogAnnotation-s got same behavior,  documentation/example:  [slf4j logger annotations](https://github.com/zio/zio-logging/blob/master/docs/overview/index.md#slf4j-logger-name-and-annotations)

Ellzord:
LGTM

Issue #445, opened 2022-04-20
SLF4J provide logging backend that stores everything in MDC 


domdorn:
This corresponds to https://github.com/zio/zio/issues/6673
I'm using SLF4J and the LogStashEncoder. I'm trying to put all fields/spans into MDCFields and just keep the real message in the message field. The only way I was able to accomplish this was by creating my own implementation of the SLF4J backend like 

```scala

  private def slf4jLogger(
                           rootLoggerName: ZTraceElement => String,
                           logLevel: LogLevel,
                           format: LogFormat
                         ): ZLogger[String, Unit] =
    new ZLogger[String, Unit] {
      val formatLogger: ZLogger[String, Option[String]] =
        format.toLogger.filterLogLevel(_ >= logLevel)

      override def apply(
                          trace: ZTraceElement,
                          fiberId: FiberId,
                          logLevel: LogLevel,
                          message: () => String,
                          cause: Cause[Any],
                          context: Map[FiberRef[_], Any],
                          spans: List[LogSpan],
                          annotations: Map[String, String]
                        ): Unit =
        formatLogger(trace, fiberId, logLevel, message, cause, context, spans, annotations).foreach { message =>
          val slf4jLogger = LoggerFactory.getLogger(rootLoggerName(trace))

          val now = System.currentTimeMillis()
          val newAnnotations = spans.map(span => span.label -> (now-span.startTime).toString).toMap ++ annotations ++ Map(
            "fiberId" -> fiberId.threadName.toString
          )

          val previous =
            if (newAnnotations.nonEmpty) {
              val previous =
                Some(Option(MDC.getCopyOfContextMap).getOrElse(java.util.Collections.emptyMap[String, String]()))
              MDC.setContextMap(newAnnotations.asJava)
              previous
            } else None

          try logLevel match {
            case LogLevel.All     => if (slf4jLogger.isTraceEnabled) slf4jLogger.trace(message)
            case LogLevel.Debug   => if (slf4jLogger.isDebugEnabled) slf4jLogger.debug(message)
            case LogLevel.Info    => if (slf4jLogger.isInfoEnabled) slf4jLogger.info(message)
            case LogLevel.Warning => if (slf4jLogger.isWarnEnabled) slf4jLogger.warn(message)
            case LogLevel.Error   => if (slf4jLogger.isErrorEnabled) slf4jLogger.error(message)
            case LogLevel.Fatal   => if (slf4jLogger.isErrorEnabled) slf4jLogger.error(message)
[...]

PR #479, opened 2022-07-25
slf4j LogAppender

justcoon:
slf4j log appender 

- key-values - `LogAppender.appendKeyValue` - logged into MDC
- cause - `LogAppender.appendCause` - as throwable, `Cause` transformed into `FiberFailure`
- slf4j logger name extracted from `Trace`

open question -
- do we need `rootLogLevel` in `private def slf4jLogger(rootLoggerName: Trace => String, rootLogLevel: LogLevel, format: LogFormat): ZLogger[String, Unit]` ? - in general, log level setup is in specific logger configuration
- `LogFormat` combinators/composition - in current design, it is not clear, where specific LogFormat part will be "inserted" (message, MDC, cause ...), there is a question, if it should not be more clear in LogFormat itself
- should we allow custom definition for log logger name (maybe by annotation as it was in version 1) ? 

related to: #485, #453, #445


987Nabil:
@justcoon We impl. our service based on the `rootLogLevel` in private def `slf4jLogger(rootLoggerName: Trace => String, rootLogLevel: LogLevel, format: LogFormat): ZLogger[String, Unit]`. This was the way I found to make different logger settings based on the logger name work.

For example
```xml
<logger name="access" level="INFO" additivity="false">
  <appender-ref ref="ACCESS_CONSOLE_LOGGER"/>
</logger>
```

And the default logic you impl. is not useful
```scala
  def getLoggerName(default: String = "zio-slf4j-logger"): Trace => String =
    _ match {
      case Trace(location, _, _) =>
        val last = location.lastIndexOf(".")
        if (last > 0) {
          location.substring(0, last)
        } else location

      case _ => default
    }
```

Is there another way to do this? If not, I'd like to have this working (soon) 🙂

987Nabil:
@justcoon I guess this is also the reason, why updating to 2.0.1 broke our slf4j + logstash json logging?
Any plan when this would be fixed? Or do you need some help?

And to answer your last question: Yes we definitely need the ability to set logger names. In the 2xRC version we did this with the `defaultLoggerName: Trace => String`

justcoon:
hello @987Nabil, 

in relation to 

> 2.0.1 broke our slf4j + logstash json logging

do you mean update from 2.0.0 -> 2.0.1? If yes, maybe it will be good if you open some ticket with details (like logstash configuration, zio logging slf4j setup - log format ..., code snippet, if possible)

in case of logger name - `SLF4J.getLoggerName`
basically if there is trace `zio.logging.example.LivePingService.ping(PingService.scala:22)`, then logger name is `zio.logging.example.LivePingService` - service class name

in my opinion, this is reasonable default logger name resolution, based on trace
but at the other side, i think, have option to say, what is logger name, can be very usefull

core/shared/src/main/scala/zio/logging/internal/LogAppender.scala
```diff
@@ -46,16 +46,11 @@ private[logging] trait LogAppender { self =>
   /**
    * Appends a key/value string pair to the log.
    */
-  final def appendKeyValue(key: String, value: String): Unit = appendKeyValue(key, _.appendText(value))
-
-  /**
-   * Appends a key/value pair, with the value it created with the log appender.
-   */
-  final def appendKeyValue(key: String, appendValue: LogAppender => Unit): Unit = {
+  def appendKeyValue(key: String, value: String): Unit = {
```
jdegoes, line 49: Let's not modify the `appendKeyValue`. Instead, generate an anonymous `LogAppender` that will simply produce string for any possible combination of operations.

987Nabil:
> hello @987Nabil,
> 
> in relation to
> 
> > 2.0.1 broke our slf4j + logstash json logging
> 
> do you mean update from 2.0.0 -> 2.0.1? If yes, maybe it will be good if you open some ticket with details (like logstash configuration, zio logging slf4j setup - log format ..., code snippet, if possible)
I'll do.
> in case of logger name - `SLF4J.getLoggerName` basically if there is trace `zio.logging.example.LivePingService.ping(PingService.scala:22)`, then logger name is `zio.logging.example.LivePingService` - service class name
> 
> in my opinion, this is reasonable default logger name resolution, based on trace but at the other side, i think, have option to say, what is logger name, can be very usefull
Maybe it is a reasonable default, but in my opinion having it configurable is a must. There are architecture guidelines out there in the wild, that require special logger names for access logging for example. One reason for something like this can be, that messages with a certain name get treated in your logging tool based on logger names.
Also it is available in every (JVM) logging lib out there and I do not see any reason to break this pattern.

987Nabil:
@justcoon I opened #489 and would not be surprised, if it is related.

justcoon:
latest changes
* previously removed function  `def appendKeyValue(key: String, appendValue: LogAppender => Unit)` of `LogAppender`- returned back
* added `SLF4J.loggerName` aspect, for custom slf4j logger name
* layers, with `LogLevel` are deprecated 
* there are scala-2 tests only, for slf4j, as there is issue with `Trace` in scala-3 (https://github.com/zio/zio/pull/7177)
* updated documentation and examples

core/shared/src/main/scala/zio/logging/internal/LogAppender.scala
justcoon: @jdegoes: previously removed function def appendKeyValue(key: String, appendValue: LogAppender => Unit) of LogAppender- returned back

jdegoes (approved):

jdegoes (approved):

Issue #322, opened 2021-09-07
Support(bridge) for java.lang.System.Logger


alek-w:

jdegoes:
This should be implemented as a `RuntimeConfigAspect` which calls `RuntimeConfigAspect.addLogger`, following the pattern laid down in `zio.logging` package object.

Any volunteers for this ticket?

justcoon:
i am working on impl. ...

987Nabil:
@justcoon this can be closed or?

justcoon:
hi @987Nabil , i think it may be closed, will do it

PR #493, opened 2022-08-18
LogAppender for java platform/system logger

justcoon:
similar to `slf4j` backend implementation

default behavior:
- all annotations are placed at the beginning of log message
- cause - LogAppender.appendCause - as throwable, Cause transformed into FiberFailure
- logger name extracted from `Trace`, added `JPL.loggerName` zio LogAnnotation for custom logger name
- updated documentation and examples

open questions:
- `slf4j` and this `jpl` implementation got custom logger name aspect, should not we do common one?
-  by default both are  appending all LogAnnotations-s, including logger name annotation, there is question if this annotation should not be excluded in default setup, because if logger name is extracted from Trace, there is not logger name annotation

related to  #322 (from this ticket it is not clear if this was requirement for backend implementation like slf4j or it should be something similar to slf4j-bridge)


justcoon:
NOTE: 
CI setup needs to be updated, as java platform/system logger was introduced at [`JDK9`](https://openjdk.org/jeps/264)

jpl/src/main/scala/zio/logging/backend/JPL.scala
```diff
@@ -0,0 +1,149 @@
...
+
+  val logFormatDefault: LogFormat = LogFormat.allAnnotations + LogFormat.line + LogFormat.cause
+
+  def loggerName(value: String): ZIOAspect[Nothing, Any, Nothing, Any, Nothing, Any] =
+    new ZIOAspect[Nothing, Any, Nothing, Any, Nothing, Any] {
+      def apply[R, E, A](zio: ZIO[R, E, A])(implicit trace: Trace): ZIO[R, E, A] =
+        ZIO.logAnnotate(loggerNameKey, value)(zio)
+    }
+
+  def getLoggerName(default: String = "zio-jpl-logger"): Trace => String =
```
jdegoes, line 43: This should be hidden / made private.

jpl/src/main/scala/zio/logging/backend/JPL.scala
```diff
@@ -0,0 +1,149 @@
...
+    loggerName: Trace => String
+  ): ZLayer[Any, Nothing, Unit] =
+    Runtime.addLogger(jplLogger(format, loggerName))
+
+  def jpl(
+    format: LogFormat
+  ): ZLayer[Any, Nothing, Unit] =
+    jpl(format, getLoggerName())
+
+  def jpl: ZLayer[Any, Nothing, Unit] =
```
jdegoes, line 112: This can be a `val`.

jdegoes (changes requested):
Just minor tweaks, and then good to merge!

justcoon:
@jdegoes - comments solved

jdegoes (approved):

PR #505, opened 2022-09-10
Handle colored log format in json appender

davidlar:


core/shared/src/main/scala/zio/logging/internal/LogAppender.scala
```diff
@@ -171,9 +173,16 @@ private[logging] object LogAppender {
...
       val cleanedTextContent = {
-        // Do a little cleanup to handle default log formats (quoted and spaced)
+        // Do a little cleanup to handle default log formats (quoted, colored and spaced)
         if (current.textContent.startsWith("\"") && current.textContent.endsWith("\""))
           current.textContent = current.textContent.drop(1).dropRight(1)
+
+        if (AnsiColorRegex.findFirstMatchIn(current.textContent).nonEmpty) {
+          val cleaned = AnsiColorRegex.replaceAllIn(current.textContent, "")
+          current.textContent.clear()
+          current.textContent.appendAll(cleaned)
```
justcoon, line 183: to avoid unnecessary clear and appendAll

i think we can add  `if(cleaned.size != current.texContent.size)`
which i think should indicate if there was some replacement ...

davidlar: Actually that's why I added `if (AnsiColorRegex.findFirstMatchIn(current.textContent).nonEmpty)`
The thought is that we first check if there is any ansi escape character.
If not, (the most common case) we will do nothing. If there is at least one, we'll run the replaceAll and then clear() + appendAll().
So if we're not rewriting the current if statement, the suggested, `if(cleaned.size != current.texContent.size)` will make no difference.

justcoon (approved):

justcoon:
@davidlar can you rebase this branch onto master, then i think we can merge this PR, thanks

davidlar:
I hope I managed to rebase it. I nearly screwed up... :smiley:

justcoon (approved):

justcoon:

Issue #509, opened 2022-09-15
log filtering improvements


justcoon:
current version of `ZLogger` (zio 2.0.2) and also other loggers implementations, in general, they are supporting just log filtering by `LogLevel` (`Zlogger.filterLogLevel`)

in general log backends implementations like logback, log4j also supporting by configuration "filtering" by package/Logger name

we should have support for something like that also in `ZLogger`,

previous version of zio-logging for zio 1 had [LogFilter](https://github.com/zio/zio-logging/blob/series%2F1.x/core/shared/src/main/scala/zio/logging/LogFiltering.scala) implementation, which allowed such filtering

PR #512, opened 2022-09-15
Log filtering improvements

justcoon:
Fixes: #509 

current version of ZLogger (zio 2.0.2) and also other loggers implementations, in general, they are supporting just log filtering by LogLevel (Zlogger.filterLogLevel)

log backends implementations like logback, log4j also supporting "filtering" by package/Logger name (in configuration)

we should have support for something like that also in `ZLogger`,

LogFiltering support based on implementation from zio-logging  zio 1 version -  [LogFiltering](https://github.com/zio/zio-logging/blob/series%2F1.x/core/shared/src/main/scala/zio/logging/LogFiltering.scala) (thanks to @vigoo)

open problems and questions:
- now in zio.logging.package, we have a lot of specific functions for logging layers, i think we should find way, how to do it more customisable and composable (have option to customise `ZLogger`, logger output and then create layer from that)


core/shared/src/main/scala/zio/logging/LogFilter.scala
```diff
@@ -0,0 +1,300 @@
...
+
+import zio.stm.{ TMap, ZSTM }
+import zio.{ Cause, FiberId, FiberRefs, LogLevel, LogSpan, Trace, ZLogger }
+
+import scala.annotation.tailrec
+
+/**
+ * A `LogFilter` represents function/conditions for log filtering
+ */
+trait LogFilter { self =>
```
justcoon, line 11: def cacheWith[A](canonicalize: LogGroup[A]): LogFilter

final case class LogGroup[A](group: (Trace, LogLevel, FiberRefs, Map[String, String]) => A)

justcoon: def filter(logger: ZLogger): ZLogger

core/shared/src/main/scala/zio/logging/LogFilter.scala
```diff
@@ -0,0 +1,375 @@
...
+
+  /**
+   * Returns a new log filter which satisfy result of this and given log filter
+   */
+  final def &&[M <: Message](other: LogFilter[M]): LogFilter[M] = and(other)
+
+  /**
+   * The alphanumeric version of the `||` operator.
+   */
+  final def or[M <: Message](other: LogFilter[M]): LogFilter[M] =
```
jdegoes, line 82: Be sure to alphabetize all these methods!

core/shared/src/main/scala/zio/logging/LogFilter.scala
```diff
@@ -0,0 +1,375 @@
...
+
+  /**
+   * Returns a new log filter with negated result
+   */
+  final def unary_! : LogFilter[Message] = self.not
+
+  /**
+   * Returns a new log filter with cached results based on given log group
+   */
+  final def cacheWith[A](group: LogGroup[A]): LogFilter[Message] =
```
jdegoes, line 131:
```suggestion
  final def cachedBy[A](group: LogGroup[A]): LogFilter[Message] =
```

core/shared/src/main/scala/zio/logging/LogFilter.scala
```diff
@@ -0,0 +1,375 @@
...
+   * @param rootLevel Default log level
+   * @param group Log group
+   * @param matcher Mather for log group and groupings
+   * @param groupings Log levels definitions
+   * @return A filter for log filtering based on given groups
+   */
+  def logLevelByGroup[A](
+    rootLevel: LogLevel,
+    group: LogGroup[A],
+    matcher: (A, A) => Boolean,
```
jdegoes, line 226: I am not sure, but could it make sense to push the `(A, A) => Boolean` into `LogGroup` as a sort of `isEquivalent`? By default it could be `l == r`, but then you could override that to have specific logic for equivalence of two `A` groups.

In fact, maybe we should have: `final case class LogGroupEquivalence[A](equivalent: (A, A) => Boolean)` just so we have a place to put constructors. Then we could consider putting that into `LogGroup` too.

core/shared/src/main/scala/zio/logging/LogFilter.scala
```diff
@@ -0,0 +1,375 @@
...
+   * @return A filter for log filtering based on log level and name
+   */
+  def logLevelByGroup(
+    rootLevel: LogLevel,
+    group: LogGroup[String],
+    mappings: (String, LogLevel)*
+  ): LogFilter[Any] = {
+    val mappingsSorted = mappings.map(splitNameByDotAndLevel.tupled).sorted(nameLevelOrdering)
+    val nameGroup      = group.map(splitNameByDot)
+    val matcher        = (loggerName: List[String], groupName: List[String]) => loggerName.startsWith(groupName)
```
jdegoes, line 293: This example is pretty compelling for why `matcher` should be inside `LogGroup`: because anytime you construct `LogGroup[String]` based on log name, you probably want startsWith equivalence.

core/shared/src/main/scala/zio/logging/LogFilter.scala
```diff
@@ -0,0 +1,375 @@
...
+      val (xName, xLevel) = x
+      val (yName, yLevel) = y
+      val r               = compareNames(xName, yName)
+      if (r == 0) { // paths are same
+        xLevel.ordinal - yLevel.ordinal
+      } else r
+    }
+  }
+
+  implicit class ZLoggerLogFilterOps[M, O](logger: zio.ZLogger[M, O]) {
```
jdegoes, line 367: Since this won't be found (by implicit search), I'd delete it.

jdegoes (changes requested):
Looks really nice! Just a few small suggestions here and there.

core/shared/src/main/scala/zio/logging/LogGroupEquivalence.scala
```diff
@@ -0,0 +1,12 @@
+package zio.logging
+
+final case class LogGroupEquivalence[A](equivalent: (A, A) => Boolean)
```
jdegoes, line 3: In order to put some force behind the word `Equivalence`, we could make it a law that `equivalent` is commutative: if `a1` is equivalent to `a2`, then `a2` is equivalent to `a1`.

On the other hand, if this would cause a problem for log groups, then we should change the name to `LogGroupRelation` andd the name of the function to `matchesGroup` rather than `equivalent`.

jdegoes, line 3:
```suggestion
final case class LogGroupEquivalence[-A](equivalent: (A, A) => Boolean)
```

We can easily make this contravariant in `A`.

core/shared/src/main/scala/zio/logging/LogGroup.scala
```diff
@@ -0,0 +1,92 @@
...
+import zio.{ FiberRefs, LogLevel, Trace, Zippable }
+
+trait LogGroup[A] { self =>
+
+  def apply(
+    trace: Trace,
+    logLevel: LogLevel,
+    context: FiberRefs,
+    annotations: Map[String, String]
+  ): A
```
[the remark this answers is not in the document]
jdegoes: Probably best to add everything we can, just so we don't have to change it later. We can always ignore parameters that we are not using!

core/shared/src/main/scala/zio/logging/LogGroupEquivalence.scala
jdegoes, line 3: A data type like this naturally supports: `&&`, `||`, and `!`, all derived from Boolean.

core/shared/src/main/scala/zio/logging/LogGroup.scala
```diff
@@ -0,0 +1,92 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package zio.logging
+
+import zio.{ FiberRefs, LogLevel, Trace, Zippable }
+
+trait LogGroup[A] { self =>
```
jdegoes, line 20: This is where you could explore adding:

```suggestion
trait LogGroup[A] { self =>
  def relation: LogGroupEquivalence[A]
```

Thus, every log group would both provide a group mapping to some type `A`, as well as a relation on values of type `A`. This sort of makes sense because in order to use the `A`, we need some capability on it (the relation), so bundling it directly inside is similar to how `FiberRef` avoids type classes by bundling fork/join operations inside it.

This slightly complicates `++`, which must take an `Unzippable`, so it can extract out the pieces and provide them to the relation.

core/shared/src/main/scala/zio/logging/LogGroup.scala
```diff
@@ -0,0 +1,92 @@
...
+      annotations: Map[String, String]
+    ): C =
+      f(self(trace, logLevel, context, annotations), other(trace, logLevel, context, annotations))
+  }
+
+}
+
+object LogGroup {
+
+  def make(loggerNameExtractor: LoggerNameExtractor): LogGroup[String] =
```
jdegoes, line 71:
```suggestion
  def fromLoggerNameExtractor(loggerNameExtractor: LoggerNameExtractor): LogGroup[String] =
```

core/shared/src/main/scala/zio/logging/LogGroup.scala
```diff
@@ -0,0 +1,92 @@
...
+      context: FiberRefs,
+      annotations: Map[String, String]
+    ): C =
+      f(self(trace, logLevel, context, annotations), other(trace, logLevel, context, annotations))
+  }
+
+}
+
+object LogGroup {
+
```
jdegoes, line 70: I'd add:

```suggestion
def apply[A](group: (Trace, LogLevel, FiberRefs, Map[String, String] => A, relation: LogGroupEquivalence[A]): LogGroup[A] = ...
```

core/shared/src/main/scala/zio/logging/LoggerNameExtractor.scala
```diff
@@ -0,0 +1,84 @@
...
+
+import zio.{ FiberRefs, Trace }
+
+trait LoggerNameExtractor { self =>
+
+  def apply(
+    trace: Trace,
+    context: FiberRefs,
+    annotations: Map[String, String]
+  ): String
```
jdegoes, line 26: If you change this `String` to `Option[String]`, then you can define `||` (or else) on `LoggerNameExtractor`, which will let you implement `annotationOrTrace` as `annotation || trace` (a more compositional definition).

core/shared/src/main/scala/zio/logging/LogGroup.scala
```diff
@@ -0,0 +1,92 @@
...
+  /**
+   * Log group by level
+   */
+  val level: LogGroup[LogLevel] =
```
jdegoes:
```suggestion
  val logLevel: LogGroup[LogLevel] =
```

core/shared/src/main/scala/zio/logging/LogFilter.scala
```diff
@@ -0,0 +1,345 @@
...
+/**
+ * A `LogFilter` represents function/conditions for log filtering
+ */
+trait LogFilter[-Message] { self =>
+
+  type Value
+
+  private[logging] def group: LogGroup[Message, Value]
+
+  private[logging] def predicate(value: Value): Boolean
```
jdegoes, line 31: I would make these two methods public, or (the alternative) make `LogFilter` sealed. Or maybe even both. Given the constructor allows you to make a log filter given a group and predicate, there is no real harm in exposing them here.

core/shared/src/main/scala/zio/logging/LogFilter.scala
```diff
@@ -0,0 +1,345 @@
...
+
+  /**
+   * Returns a new log filter with negated result
+   */
+  final def unary_! : LogFilter[Message] = self.not
+
+  /**
+   * Returns a version of logger that only logs messages when this filter is satisfied
+   */
+  def filter[M <: Message, O](logger: zio.ZLogger[M, O]): zio.ZLogger[M, Option[O]] =
```
jdegoes, line 118: Be sure to alphabetize all these methods.

jdegoes (dismissed):

jdegoes (approved):

Issue #520, opened 2022-09-20
Support slf4j markers in SLF4J backend


jeejeeone:
Support slf4j `org.slf4j.Marker` in SLF4J backend. Currently slf4j Marker API cannot be used with zio-logging-slf4j SLF4J backend. This is inconvenient in some use cases, for example when using an ECS Layout with `includeMarkers`.

For example write info log in current implementation:
`slf4jLogger.info(message.toString, throwable)`

It should be possible to pass `org.slf4j.Marker` to SLF4J appender, used slf4j api methods should use marker if applicable.

Info log with marker:
`slf4jLogger.info(marker, message.toString, throwable)`

Should also be applied to is--Enabled methods:
`slf4jLogger.isTraceEnabled(marker)`

justcoon:
hi @jeejeeone, I made draft PR #521, please check it and let me know, what do you think about that ...

jeejeeone:
@justcoon Hello, looks good to me! My use case with ECS layout and includeMarkers works ok and the API is nice.

justcoon:
@jeejeeone , thanks you for answer

Just for clarity (i mentioned there 2 variants), for your use case:
2. variant where, there is log marker name provided
`val confidentialMarker = SLF4J.logMarkerName("CONFIDENTIAL")`

Do you think that variant? I like it more, it look nicer, but may be limitation for some very specific use cases, where somebody  like to provide custom Marker implementation

jeejeeone:
In my previous reply I was referring to the second variant :-) Yeah, it's nice, not sure about custom Marker implementations.

PR #521, opened 2022-09-20
slf4j markers

justcoon:
slf4j log marker 
slf4j default log format with excluded annotations for logger name and log marker
jpl default log format with excluded annotation for logger name
LogFormat annotations with exclude by keys

 

Fixes: #520

### DIFF
--- a/core/shared/src/main/scala/zio/logging/LogFormat.scala
+++ b/core/shared/src/main/scala/zio/logging/LogFormat.scala
@@ -234,9 +234,7 @@ object LogFormat {
     LogFormat.make { (builder, _, _, _, _, _, fiberRefs, _, _) =>
       fiberRefs
         .get(logContext)
-        .foreach { anyRef =>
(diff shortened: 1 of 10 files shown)


core/shared/src/main/scala/zio/logging/LogFormat.scala
```diff
@@ -248,26 +246,36 @@ object LogFormat {
...
+        if (!excludeKeys.contains(key)) {
+          builder.appendKeyValue(key, value)
+        }
       }
     }
 
-  def logAnnotations: LogFormat =
+  def logAnnotations: LogFormat = logAnnotations(Set.empty)
+
+  def logAnnotations(excludeKeys: Set[String]): LogFormat =
```
jdegoes, line 262: I would make `final case class ExcludeKeys(keys: Set[String])` just so we can at some point also have include keys and not break code.

jdegoes (approved):

Issue #524, opened 2022-09-29
Shall we add support for scala 2.13.9?


jiahuili430:
`CVE-2022-36944` exists in scala 2.13.0 ~ 2.13.8, see https://nvd.nist.gov/vuln/detail/CVE-2022-36944
Are we going to add support for scala 2.13.9?

justcoon:
@jiahuili430 thank you for report,
tried make update, however there is bug https://github.com/scala/bug/issues/12650

PR #525, opened 2022-09-29
scala 2.13.10

justcoon:
fixes: #524


justcoon:
mimaChecks broken https://github.com/scala/bug/issues/12650

PR #529, opened 2022-10-06
Give proper logger name to SLF4J bridge

guersam:
Current `org.slf4j.impl.ZioLogger` doesn't implement `Logger#getName` that causes a silent failure when using Netty, which rejects slf4j loggers without name.

### DIFF
--- a/slf4j-bridge/src/main/scala-2/org/slf4j/impl/ZioLogger.scala
+++ b/slf4j-bridge/src/main/scala-2/org/slf4j/impl/ZioLogger.scala
@@ -4,6 +4,9 @@ import org.slf4j.helpers.{ MarkerIgnoringBase, MessageFormatter }
 import zio.{ Cause, ZIO }
 
 class ZioLogger(name: String, factory: ZioLoggerFactory) extends MarkerIgnoringBase {
+
+  override def getName: String = name
+
   private def run(f: ZIO[Any, Nothing, Unit]): Unit =
     factory.run {
       ZIO.logSpan(name)(f)
--- a/slf4j-bridge/src/main/scala-2/org/slf4j/impl/ZioLogger.scala
+++ b/slf4j-bridge/src/main/scala-2/org/slf4j/impl/ZioLogger.scala
@@ -1,11 +1,9 @@
 package org.slf4j.impl
 
-import org.slf4j.helpers.{ MarkerIgnoringBase, MessageFormatter }
+import org.slf4j.helpers.{ MessageFormatter, ZioLoggerBase }
 import zio.{ Cause, ZIO }
 
-class ZioLogger(name: String, factory: ZioLoggerFactory) extends MarkerIgnoringBase {
-
-  override def getName: String = name
+class ZioLogger(name: String, factory: ZioLoggerFactory) extends ZioLoggerBase(name) {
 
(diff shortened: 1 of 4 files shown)


slf4j-bridge/src/main/scala-2/org/slf4j/impl/ZioLogger.scala
```diff
@@ -4,6 +4,9 @@ import org.slf4j.helpers.{ MarkerIgnoringBase, MessageFormatter }
 import zio.{ Cause, ZIO }
 
 class ZioLogger(name: String, factory: ZioLoggerFactory) extends MarkerIgnoringBase {
+
+  override def getName: String = name
```
guersam, line 8: There's already `protected String name` in `org.slf4j.helpers.NamedLoggerBase` but Scala doesn't allow implementing it with `override val`, so that I directly overrode `getName` here.

slf4j-bridge/src/test/scala/zio/logging/slf4j/bridge/Slf4jBridgeSpec.scala
```diff
@@ -83,6 +83,11 @@ object Slf4jBridgeSpec extends ZIOSpecDefault {
...
         )
-      }.provide(Slf4jBridge.initialize)
-    )
+      },
+      test("Implements Logger#getName") {
+        for {
+          logger <- ZIO.attempt(org.slf4j.LoggerFactory.getLogger("zio.test.logger"))
+        } yield assertTrue(logger.getName == "zio.test.logger")
+      }
+    ).provide(Slf4jBridge.initialize) @@ TestAspect.sequential
```
guersam, line 92: The first case fails when I change `proved` to `provideShared` or remove `TestAspect.sequential`. Do you have any idea of the reason?

slf4j-bridge/src/main/scala-2/org/slf4j/impl/ZioLogger.scala
justcoon: i think also something like:

```scala
class ZioLogger(n: String, factory: ZioLoggerFactory) extends MarkerIgnoringBase {
  this.name = n
  
}
```

could work

justcoon (approved):
thank you for your contribution

slf4j-bridge/src/main/scala-2/org/slf4j/impl/ZioLogger.scala
guersam: Yes that will work. It's been too long since I wrote mutable Scala code :stuck_out_tongue: 

Would you prefer this to introducing intermediate abstract class as in 5b719d1a986108daa0c9a9585777f5a7896f71b8?

justcoon: solution with ZioLoggerBase is probably nicer, I am ok with that

slf4j-bridge/src/test/scala/zio/logging/slf4j/bridge/Slf4jBridgeSpec.scala
guersam: It seems like expected behavior of `ZTestLogger.default`.

PR #556, opened 2022-11-11
Fix StaticMarkerBinder implementation in slf4j-bridge

n3ziniuka5:
I noticed this issue when using ZIO with Akka and attempted to route Akka logs via the `slf4j-bridge`. Akka calls `MarkerFactory.getIMarkerFactory`, which throws an exception in the existing implementation.

### DIFF
--- a/slf4j-bridge/src/main/scala/org/slf4j/impl/StaticMarkerBinder.scala
+++ b/slf4j-bridge/src/main/scala/org/slf4j/impl/StaticMarkerBinder.scala
@@ -5,11 +5,14 @@ import org.slf4j.helpers.BasicMarkerFactory
 import org.slf4j.spi.MarkerFactoryBinder
 
 class StaticMarkerBinder extends MarkerFactoryBinder {
-  override def getMarkerFactory: IMarkerFactory = StaticMarkerBinder.singleton
-  override def getMarkerFactoryClassStr: String = StaticMarkerBinder.className
+  private val markerFactory = new BasicMarkerFactory
+
+  override def getMarkerFactory: IMarkerFactory = markerFactory
+  override def getMarkerFactoryClassStr: String = classOf[BasicMarkerFactory].getName
 }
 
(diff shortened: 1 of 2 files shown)


slf4j-bridge/src/main/scala/org/slf4j/impl/StaticMarkerBinder.scala
```diff
@@ -5,11 +5,14 @@ import org.slf4j.helpers.BasicMarkerFactory
...
+  override def getMarkerFactory: IMarkerFactory = markerFactory
+  override def getMarkerFactoryClassStr: String = classOf[BasicMarkerFactory].getName
 }
 
 object StaticMarkerBinder {
-  private val singleton = new BasicMarkerFactory
-  private val className = classOf[BasicMarkerFactory].getName
+  private val singleton = new StaticMarkerBinder
+
+  def getSingleton: StaticMarkerBinder = singleton
```
justcoon, line 17: is this needed just for testing? or it is used also in other way?

n3ziniuka5: This is required by `MarkerFactory.getIMarkerFactory`, which does:
```java
try {
    return StaticMarkerBinder.getSingleton().getMarkerFactory();
} catch (NoSuchMethodError nsme) {
    // binding is probably a version of SLF4J older than 1.7.14
    return StaticMarkerBinder.SINGLETON.getMarkerFactory();
}
```

The existing implementation didn't have `getSingleton` and the `singleton` that was defined previously was pointing to the factory instead of the binder as expected by `slf4j`. So an exception is thrown.

justcoon: yeah, noticed that later, but thanks for clarification

justcoon (approved):

PR #590, opened 2022-12-20
Log Metrics

justcoon:
one of the zio logging related questions was: how to add counter/metric for error logs

As ZIO core supporting multiple loggers, this logging metrics collector is implemented as specific `ZLogger` 
which is responsible just for collecting metrics of all logs - `ZIO.log*` functions. (thanks to @vigoo for this idea)

metrics in form of counter for each log level, metric name:

zio_logger_${level.label.toLowerCase}_total
for example for LogLevel.Error: zio_logger_error_total

alternative solution for counter definition could be:

common metric name, for example zio_logger_log_total
log level in metric tags

### DIFF
--- a/core/shared/src/main/scala/zio/logging/package.scala
+++ b/core/shared/src/main/scala/zio/logging/package.scala
@@ -15,6 +15,8 @@
  */
 package zio
 
+import zio.metrics.Metric
+
 import java.io.PrintStream
 import java.nio.charset.{ Charset, StandardCharsets }
 import java.nio.file.Path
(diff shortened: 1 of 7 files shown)


core/shared/src/main/scala/zio/logging/package.scala
```diff
@@ -299,4 +313,25 @@ package object logging {
...
+    ): Unit = {
+      loggedTotalMetrics.get(logLevel).foreach { counter =>
+        val tags = context.get(FiberRef.currentTags).getOrElse(Set.empty)
+        counter.unsafe.update(1, tags)(Unsafe.unsafe)
+      }
+      ()
+    }
+  }
+
+  val metrics: ZLayer[Any, Nothing, Unit] = Runtime.addLogger(metricLogger)
```
jdegoes, line 336: Maybe `logMetrics` to avoid clashing with identical names in scope?

jdegoes (dismissed):

core/shared/src/main/scala/zio/logging/package.scala
justcoon: ok, that is good idea

jdegoes (approved):

PR #597, opened 2022-12-27
Idiomatic log metrics with customizable naming

987Nabil:
The current impl. creates many metrics, which is not idiomatic. They are all the same metric with different notions. Therefore I propose to use labels instead with only one counter.
Also I think being forced to use a certain metric name is not good, since orgs might enforce naming conventions.

I remember, that we have the option for custom naming in other zio metrics too. I think it was zio-http.

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -3,31 +3,40 @@ id: metrics
 title: "Log Metrics"
 ---
 
-Log metrics collecting metrics related to ZIO logging (all `ZIO.log*` functions). 
-As ZIO core supporting multiple loggers, this logging metrics collector is implemented as specific `ZLogger` 
+Log metrics collecting metrics related to ZIO logging (all `ZIO.log*` functions).
+As ZIO core supporting multiple loggers, this logging metrics collector is implemented as specific `ZLogger`
 which is responsible just for collecting metrics of all logs - `ZIO.log*` functions.
 
-Metrics layer:
+The Metrics layer
 
 ```scala
 val layer = zio.logging.logMetrics
 ```
 
-Metrics:
-* zio_logger_all_total - logs count for `LogLevel.All`
-* zio_logger_fatal_total - logs count for `LogLevel.Fatal`
-* zio_logger_error_total - logs count for `LogLevel.Error`
-* zio_logger_warn_total - logs count for `LogLevel.Warning`
-* zio_logger_info_total - logs count for `LogLevel.Info`
-* zio_logger_debug_total - logs count for `LogLevel.Debug`
-* zio_logger_trace_total - logs count for `LogLevel.Trace`
-* zio_logger_none_total - logs count for `LogLevel.None`
+will add a default metric named `zio_logger_total` with the label `level` which will be
+incremented for each log message with the value of `level` being the corresponding log level in lower case.
 
+Metrics:
+
+* `LogLevel.All` -> `all`
+* `LogLevel.Fatal` -> `fatal`
+* `LogLevel.Error` -> `error`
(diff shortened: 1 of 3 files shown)


justcoon:
hi @987Nabil,

thank you for contribution,

in my opinion is hard to say what is idiomatic in this case, 

in various systems/frameworks/applications both solution are used: 
* custom metrics per problem/category
* one common metric with tags

i think it it's about preferences in terms of options (of course sometimes one solution may be better then other ...)

originally i made metric per log level, to have similar style like in zio core (there are specific metrics, for zio_fiber_started,  zio_fiber_successes, zio_fiber_failures), 

basically in original PR (#590) i mentioned also alternative solution which is same like here in your PR

but yes, it is easier to change naming in case of common metric with tags

personally I am ok with your proposal, more like there may be question in relation to backward compatibility, because this change is not backward compatible

docs/metrics.md
```diff
@@ -7,22 +7,31 @@ Log metrics collecting metrics related to ZIO logging (all `ZIO.log*` functions)
...
-* zio_logger_error_total - logs count for `LogLevel.Error`
-* zio_logger_warn_total - logs count for `LogLevel.Warning`
-* zio_logger_info_total - logs count for `LogLevel.Info`
-* zio_logger_debug_total - logs count for `LogLevel.Debug`
-* zio_logger_trace_total - logs count for `LogLevel.Trace`
-* zio_logger_none_total - logs count for `LogLevel.None`
+* `LogLevel.All` -> `all`
+* `LogLevel.Fatal` -> `fatal`
+* `LogLevel.Error` -> `error`
+* `LogLevel.Warning` -> `warning`
```
justcoon, line 23: label for LogLevel.Warning is WARN
then this should be:

`LogLevel.Warning` -> `warn`

987Nabil:
@justcoon I think Scala api wise it is. Only the name of the metric is a problem. But this is out now for 2 days?

All places I came a long sofar prefer labels.
And it is afaik more efficient in the metric databases, since they are optimized for it.

docs/metrics.md
```diff
@@ -7,22 +7,31 @@ Log metrics collecting metrics related to ZIO logging (all `ZIO.log*` functions)
...
+* `LogLevel.Info` -> `info`
+* `LogLevel.Debug` -> `debug`
+* `LogLevel.Trace` -> `trace`
+* `LogLevel.None` -> `none`
```
justcoon, line 27: label for LogLevel.Warning is OFF
then this should be:

`LogLevel.None` -> `off`

NOTE: here is bug also in original documentation

justcoon:
ok, that is true, that it is released just few days (and also there are holidays now)

i added some comments, also code needs to be reformatted (build failing on scalafmtCheck  - there is `sbt fix` command)

987Nabil:
Always the formatter. I am used to formatting that is applied and pushed by the pipeline. But I think this only works for private repos.
I am updating.

docs/metrics.md
```diff
@@ -49,21 +58,22 @@ object MetricsApp extends ZIOAppDefault {
...
+      _ <- Console.printLine(metricValues)
+      _ <- ZIO.logInfo("Done")
     } yield ExitCode.success)
       .provideLayer((ZLayer.succeed(MetricsConfig(200.millis)) ++ publisherLayer) >+> prometheusLayer)
 
 }
 ```
 
 Expected Console Output:
+
```
justcoon, line 76: in relation to metric changes, console output for this example probably also changed, 
i think you should run MetricsApp from examples and update it here

jdegoes (approved):

PR #613, opened 2023-01-08
Feature #612 - implement LogFilter glob-like wildcard support

### LINKED ISSUE #612: Feature request: support glob-like wildcard path matching for LogFilters
In a logback config I can do a glob-like logger-name filter:
```<logger name="org.my.**.*Layers" level="INFO"/>```

The ZIO LogFilter does not support this. It would be nice to set a log-level this way: 
```"org.my.**.*Layers" -> LogLevel.Info```

### PR DESCRIPTION
**Pull request by ThijsBroersen**
#612

I implemented `**` (any number of paths) and `*` (any (partial) path) within the existing LogFilter path structure `List[String]`. 

I think this could be optimized if the LogFilters were used in a nested map-like structure. 

Another way would be the use regex.

### DIFF
--- a/core/jvm/src/test/scala/zio/logging/LogFilterSpec.scala
+++ b/core/jvm/src/test/scala/zio/logging/LogFilterSpec.scala
@@ -85,9 +85,14 @@ object LogFilterSpec extends ZIOSpecDefault {
 
       val filter: LogFilter[String] = LogFilter.logLevelByName(
         LogLevel.Debug,
-        "a"     -> LogLevel.Info,
-        "a.b.c" -> LogLevel.Warning,
-        "e.f"   -> LogLevel.Error
+        "a"         -> LogLevel.Info,
+        "a.b.c"     -> LogLevel.Warning,
+        "e.f"       -> LogLevel.Error,
+        "k.*.m"     -> LogLevel.Trace,
+        "k2.a*c.m2" -> LogLevel.Trace,
+        "k3.a*.m3"  -> LogLevel.Trace,
+        "k4.*c.m4"  -> LogLevel.Trace,
+        "q.**.t"    -> LogLevel.Trace
       )
 
       testFilter(filter, "x.Exec.exec", LogLevel.Debug, Assertion.isTrue) &&
@@ -98,7 +103,20 @@ object LogFilterSpec extends ZIOSpecDefault {
       testFilter(filter, "a.b.c.Exec.exec", LogLevel.Info, Assertion.isFalse) &&
       testFilter(filter, "a.b.c.Exec.exec", LogLevel.Warning, Assertion.isTrue) &&
       testFilter(filter, "e.Exec.exec", LogLevel.Debug, Assertion.isTrue) &&
-      testFilter(filter, "e.f.Exec.exec", LogLevel.Debug, Assertion.isFalse)
+      testFilter(filter, "e.f.Exec.exec", LogLevel.Debug, Assertion.isFalse) &&
+      testFilter(filter, "k.l.Exec.exec", LogLevel.Debug, Assertion.isTrue) &&
+      testFilter(filter, "k.l.m.Exec.exec", LogLevel.Trace, Assertion.isTrue) &&
+      testFilter(filter, "k2.alc.m2.Exec.exec", LogLevel.Trace, Assertion.isTrue) &&
+      testFilter(filter, "k3.alc.m3.Exec.exec", LogLevel.Trace, Assertion.isTrue) &&
+      testFilter(filter, "k3.lc.m3.Exec.exec", LogLevel.Trace, Assertion.isFalse) &&
+      testFilter(filter, "k4.alc.m4.Exec.exec", LogLevel.Trace, Assertion.isTrue) &&
+      testFilter(filter, "k4.al.m4.Exec.exec", LogLevel.Trace, Assertion.isFalse) &&
+      testFilter(filter, "k.l.l.m.Exec.exec", LogLevel.Trace, Assertion.isFalse) &&
+      testFilter(filter, "k.l.l.m.Exec.exec", LogLevel.Debug, Assertion.isTrue) &&
+      testFilter(filter, "q.r.s.t.Exec.exec", LogLevel.Trace, Assertion.isTrue) &&
+      testFilter(filter, "q.r.t.Exec.exec", LogLevel.Trace, Assertion.isTrue) &&
+      testFilter(filter, "q.r.s.u.Exec.exec", LogLevel.Trace, Assertion.isFalse) &&
+      testFilter(filter, "q.r.s.u.Exec.exec", LogLevel.Debug, Assertion.isTrue)
     },
     test("log filtering by log level and name with annotation") {
 
--- a/core/jvm/src/test/scala/zio/logging/LogFilterSpec.scala
+++ b/core/jvm/src/test/scala/zio/logging/LogFilterSpec.scala
@@ -88,11 +88,11 @@ object LogFilterSpec extends ZIOSpecDefault {
         "a"         -> LogLevel.Info,
         "a.b.c"     -> LogLevel.Warning,
         "e.f"       -> LogLevel.Error,
-        "k.*.m"     -> LogLevel.Trace,
-        "k2.a*c.m2" -> LogLevel.Trace,
-        "k3.a*.m3"  -> LogLevel.Trace,
-        "k4.*c.m4"  -> LogLevel.Trace,
-        "q.**.t"    -> LogLevel.Trace
+        "k.*.m"     -> LogLevel.Info,
+        "k2.a*c.m2" -> LogLevel.Info,
+        "k3.a*.m3"  -> LogLevel.Info,
+        "k4.*c.m4"  -> LogLevel.Info,
+        "q.**.t"    -> LogLevel.Info
       )
 
       testFilter(filter, "x.Exec.exec", LogLevel.Debug, Assertion.isTrue) &&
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by ThijsBroersen** [comment]
Did some benchmarks to see if there is much performance loss.

Feature #612
```
[info] Benchmark                                           Mode  Cnt     Score       Error  Units
[info] FilterBenchmarks.cachedFilterByLogLevelAndNameLog  thrpt    3  5668.719 ± 39326.263  ops/s
[info] FilterBenchmarks.filterByLogLevelAndNameLog        thrpt    3  4649.646 ± 12339.049  ops/s
[info] FilterBenchmarks.handWrittenFilterLog              thrpt    3  3109.203 ±  2338.471  ops/s
[info] FilterBenchmarks.noFilteringLog                    thrpt    3  2270.438 ±  8116.408  ops/s```

Master
```[info] Benchmark                                           Mode  Cnt     Score       Error  Units
[info] FilterBenchmarks.cachedFilterByLogLevelAndNameLog  thrpt    3  5390.427 ± 35953.429  ops/s
[info] FilterBenchmarks.filterByLogLevelAndNameLog        thrpt    3  4323.591 ± 22497.521  ops/s
[info] FilterBenchmarks.handWrittenFilterLog              thrpt    3  2725.011 ±   348.357  ops/s
[info] FilterBenchmarks.noFilteringLog                    thrpt    3  1957.494 ±  1254.506  ops/s```

**Comment by justcoon** [comment]
thank you for contribution

**Comment by ThijsBroersen** [comment]
Hmm, I see this library still needs to support older Scala versions for which the CI tests now fail 😞

File: core/shared/src/main/scala/zio/logging/LogFilter.scala [hunk 1]
```diff
@@ -232,10 +232,60 @@ object LogFilter {
     val mappingsSorted = mappings.map(splitNameByDotAndLevel.tupled).sorted(nameLevelOrdering)
     val nameGroup      = group.map(splitNameByDot)
 
+    @tailrec
+    def compareRoutes(l: List[String], m: List[String]): Boolean = (l, m) match {
```
justcoon, line 236: not sure if there will be noticeable performance difference
but maybe it could be good to try implementation with traversing both list with one iterator each

basically in similar way like `List.startsWith` is implemented

because in this code it is look like there may be a lot of unnecessary iterations over both list

core/jvm/src/test/scala/zio/logging/LogFilterSpec.scala
```diff
@@ -98,7 +103,20 @@ object LogFilterSpec extends ZIOSpecDefault {
...
+      testFilter(filter, "k3.alc.m3.Exec.exec", LogLevel.Trace, Assertion.isTrue) &&
+      testFilter(filter, "k3.lc.m3.Exec.exec", LogLevel.Trace, Assertion.isFalse) &&
+      testFilter(filter, "k4.alc.m4.Exec.exec", LogLevel.Trace, Assertion.isTrue) &&
+      testFilter(filter, "k4.al.m4.Exec.exec", LogLevel.Trace, Assertion.isFalse) &&
+      testFilter(filter, "k.l.l.m.Exec.exec", LogLevel.Trace, Assertion.isFalse) &&
+      testFilter(filter, "k.l.l.m.Exec.exec", LogLevel.Debug, Assertion.isTrue) &&
+      testFilter(filter, "q.r.s.t.Exec.exec", LogLevel.Trace, Assertion.isTrue) &&
+      testFilter(filter, "q.r.t.Exec.exec", LogLevel.Trace, Assertion.isTrue) &&
+      testFilter(filter, "q.r.s.u.Exec.exec", LogLevel.Trace, Assertion.isFalse) &&
+      testFilter(filter, "q.r.s.u.Exec.exec", LogLevel.Debug, Assertion.isTrue)
```
justcoon, line 119: we should add test where we can see that path without mask take precedence over path with mask

etc
in filter
`"k3.a*.m3" -> LogLevel.Trace`  vs `"k3.alc.m3" -> LogLevel.Warning` 

then "k3.alc.m3.Exec.exec" should have `LogLevel.Warning`

ThijsBroersen: rewrote them

core/shared/src/main/scala/zio/logging/LogFilter.scala
ThijsBroersen: I think I have another idea. When creating the LogFilter it should already separate glob-like filters from simple ones. Then have separate implementations to avoid runtime penalties for the simple filters and the glob-like filter can then be implemented more efficient.

ThijsBroersen: Hmm, I also see the matcher argument is public so probably make it optional. Then when there is no need for glob-like filters there is also no performance loss.

ThijsBroersen: I will make changes to this PR somewhere in de next days.

justcoon:
hello @ThijsBroersen , hope you are doing well, 

are you planning to continue to work on this issue? 

thank you

ThijsBroersen:
PR
```
[info] Benchmark                                           Mode  Cnt      Score      Error  Units
[info] FilterBenchmarks.cachedFilterByLogLevelAndNameLog  thrpt    3  11606.209 ± 4034.046  ops/s
[info] FilterBenchmarks.filterByLogLevelAndNameLog        thrpt    3  11465.383 ± 7223.445  ops/s
[info] FilterBenchmarks.handWrittenFilterLog              thrpt    3   9420.996 ± 1500.501  ops/s
[info] FilterBenchmarks.noFilteringLog                    thrpt    3   8159.752 ± 1312.800  ops/s
```
Master
```
[info] Benchmark                                           Mode  Cnt      Score      Error  Units
[info] FilterBenchmarks.cachedFilterByLogLevelAndNameLog  thrpt    3  11525.222 ± 1802.478  ops/s
[info] FilterBenchmarks.filterByLogLevelAndNameLog        thrpt    3  11873.502 ± 2359.884  ops/s
[info] FilterBenchmarks.handWrittenFilterLog              thrpt    3   9522.591 ±  927.634  ops/s
[info] FilterBenchmarks.noFilteringLog                    thrpt    3   7957.346 ± 1085.727  ops/s
```

core/shared/src/main/scala/zio/logging/LogFilter.scala
```diff
@@ -371,7 +400,12 @@ object LogFilter {
         case (xFirst :: xTail, yFirst :: yTail) =>
           val r = yFirst.compareTo(xFirst)
           if (r != 0) {
-            r
+            if (xFirst.contains('*') || yFirst.contains('*')) {
+              if (Set("**", "*").contains(xFirst)) 1
+              else if (Set("**", "*").contains(yFirst)) -1
+              else
+                compareNames(xFirst.split('*').toList.filter(_.nonEmpty), yFirst.split('*').toList.filter(_.nonEmpty))
+            } else r
```
justcoon, line 408: LogFilterSpec.nameLevelOrdering should be extended to check proper ordering ...

core/shared/src/main/scala/zio/logging/LogFilter.scala
```diff
@@ -284,10 +284,39 @@ object LogFilter {
...
+          // try a regular, routesCompare or check if skipping paths (globstar pattern) results in a matching path
+          l.startsWith(m) || compareRoutes(l, m) || globStarCompare(ls, m)
+      }
+
+    @tailrec
+    def compareRoutes(l: List[String], m: List[String]): Boolean =
+      (l, m) match {
+        case (_, Nil)                                  => true
+        case (Nil, _)                                  => false
+        case (_ :: ls, "*" :: ms)                      => true && compareRoutes(ls, ms)
```
justcoon, line 302: **true &&** should not be needed

core/shared/src/main/scala/zio/logging/LogFilter.scala
```diff
@@ -284,10 +284,39 @@ object LogFilter {
...
+        case (l, "**" :: ms)                           => globStarCompare(l, ms)
+        case (lh :: ls, mh :: ms) if !mh.contains("*") =>
+          lh == mh && compareRoutes(ls, ms)
+        case (l @ (lh :: ls), m @ (mh :: ms))          =>
+          val (result, unprocessed) = mh
+            .split('*')
+            .toList
+            .foldLeft((true, lh)) { case ((matchResult, lr), p) =>
+              (matchResult && lr.containsSlice(p)) -> lr.drop(lr.indexOfSlice(p) + p.size)
```
justcoon, line 311: if `matchResult` is false `lr.drop(lr.indexOfSlice(p) + p.size` is unnecessary computation

core/shared/src/main/scala/zio/logging/LogFilter.scala
```diff
@@ -284,10 +284,39 @@ object LogFilter {
     val mappingsSorted = mappings.map(splitNameByDotAndLevel.tupled).sorted(nameLevelOrdering)
     val nameGroup      = group.map(splitNameByDot)
 
+    @tailrec
+    def globStarCompare(l: List[String], m: List[String]): Boolean =
+      (l, m) match {
+        case (_, Nil)                       => true
+        case (Nil, _)                       => false
+        case (l @ (_ :: ls), m @ (mh :: _)) =>
```
justcoon, line 292: `mh` does not look like used

core/shared/src/main/scala/zio/logging/LogFilter.scala
ThijsBroersen: add a dedicated method to check it recursive and only use only indexOfSlice to look fo a match

core/shared/src/main/scala/zio/logging/LogFilter.scala
ThijsBroersen: extended the spec and resolved an ordering issue between `*` vs `**`

justcoon (approved):

PR #624, opened 2023-01-18
Document the usage of `Slf4j.slf4j` layer

pierangeloc:
We had an issue with double logging due to the presence of 2 such layers in our application, and it took a while to discover the root cause. Such a documentation could help other devs in the same situation.


justcoon:
@pierangeloc thank you for contribution,

please run `sbt fix`,  CI failing on scalafmtCheck

thank you

pierangeloc:
@justcoon thank you for the feedback. I fixed the issue and amended the commit.

PR #630, opened 2023-01-23
added spec for traceCounter

aashish2054:
Added spec for traceCounter in MetricsSpec.scala


justcoon:
thank you for contribution

PR #633, opened 2023-01-24
added ZIO.logError in the spec

aashish2054:


justcoon:
thank you for contribution

Issue #637, opened 2023-01-28
slf4j v2  logging backend


justcoon:
slf4j v2 logging backend implementation:

similar like slf4j v1 implementation (for easier migrations), but using some of new features like key value pairs from fluent api https://www.slf4j.org/manual.html#fluent, instead of MDC which have issue (like #491), as MDC using ThreadLocal

Issue #491, opened 2022-08-18
slf4j/logback MDC context reset/rewritten with other Fiber before it is logged 


justcoon:
in case of parallel execution, it may happen, that MDC context will be reset/rewritten with other Fiber, before it is logged 

```scala
package zio.logging.backend

import zio.logging.LogAnnotation
import zio.test.Assertion._
import zio.test._
import zio.{ LogLevel, Runtime, ZIO, ZIOAspect, _ }

import java.util.UUID

object SLF4JParSpec extends ZIOSpecDefault {

  val loggerDefault: ZLayer[Any, Nothing, Unit] =
    Runtime.removeDefaultLoggers >>> SLF4J.slf4j

  val spec: Spec[Environment, Any] = suite("SLF4JParSpec")(
    test("log all annotations into MDC with custom logger name") {
      val users = Chunk.fill(2)(UUID.randomUUID())
      for {
        traceId <- ZIO.succeed(UUID.randomUUID())
        _        = TestAppender.reset()
        _       <- ZIO.foreachPar(users) { uId =>
                     {
                       ZIO.logInfo("Starting operation") *> ZIO.sleep(500.millis) *> ZIO.logInfo("Stopping operation")
                     } @@ ZIOAspect.annotated("user", uId.toString)
                   } @@ LogAnnotation.TraceId(traceId)
        _       <- ZIO.logInfo("Done")
      } yield {
        val loggerOutput = TestAppender.logOutput
        assertTrue(loggerOutput.size == 5) && assertTrue(
          loggerOutput.forall(_.loggerName == "zio.logging.backend.SLF4JParSpec")
        ) && assertTrue(loggerOutput.forall(_.logLevel == LogLevel.Info)) && assert(loggerOutput.map(_.message).toSet)(
          equalTo(
            Set(
              "Starting operation",
              "Stopping operation",
              "Starting operation",
              "Stopping operation",
              "Done"
            )
          )
        ) && assert(loggerOutput.map(_.mdc.get(LogAnnotation.TraceId.name)).sorted)(
          equalTo(
            (Chunk.fill(4)(Some(traceId.toString)) :+ None).sorted
          )
        ) && assert(loggerOutput.map(_.mdc.get("user")).sorted)(
          equalTo((users.flatMap(u => Chunk.fill(2)(Some(u.toString))) :+ None).sorted)
        )
      }
    }.provide(loggerDefault)
  ) @@ TestAspect.sequential @@ TestAspect.withLiveClock
}
```

[...]

PR #639, opened 2023-01-30
Slf4j v2 backend

justcoon:
depends on #638 

fixes: #637 

similar like slf4j v1 implementation (for easier migrations), but using some of new features like key value pairs 
instead of MDC which have issue (like https://github.com/zio/zio-logging/issues/491), as MDC using ThreadLocal

throwable/exception handling was changed to:
- when there is `Cause.Die` or `Cause.Fail` with Throwable, that throwable is logged
- otherwise cause is transformed to `zio.FiberFailure` - as it was originally in all cases

TODO:
- [markers](https://www.slf4j.org/apidocs/org/slf4j/Marker.html)
   - check support for multiple [Markers](https://www.slf4j.org/manual.html#fluent) - slf4j v1 backend is able to set just one marker, by marker name
   - check other marker related features - option to provide created Marker?


slf4j/src/main/scala/zio/logging/slf4j/SLF4J.scala
```diff
@@ -105,7 +105,11 @@ object SLF4J {
        */
       override def appendCause(cause: Cause[Any]): Unit = {
         if (!cause.isEmpty) {
-          throwable = FiberFailure(cause)
+          cause match {
```
jdegoes, line 108: Matching on a cause directly is not reliable because of nodes like or / and / stackless /e tc.

Instead, if you are trying to just "get the most important throwable", then do something like `squash` or `defects` or perhaps: `(cause.errors.collect { case t : Throwable } ++ cause.defects).head` (or its equivalent). 

Because you want to prefer an error before you look into the defects, and you want to be robust in the presence of intermediate / metadata nodes.

jdegoes (changes requested):
One quick comment to make extracting a `Throwable` from a `Cause` more robust.

slf4j/src/main/scala/zio/logging/slf4j/SLF4J.scala
justcoon: @jdegoes 

I changed that to
```scala
        if (!cause.isEmpty) {
          val maybeThrowable = (cause.failures.collect { case t: Throwable => t } ++ cause.defects).headOption
          maybeThrowable match {
            case Some(t) => throwable = t
            case None    => throwable = FiberFailure(cause)
          }
        }
```

can you check it? Thank you

jdegoes (approved):
Looks great! Please also open a ticket on ZIO to add `Cause#toThrowable` that does something similar (but more efficiently).

Issue #643, opened 2023-02-02
improve support for structured logging (Json and possibly other formats) 


justcoon:
in current implementation we have json logger
however there is not way have to do case class logging to json in structured way

there is also limitation in relation to ZIO#log* interface which only supports strings, zio core log annotation also supporting just string values
in relation to that currently it could be possible to do

```scala
object ConsoleJsonApp extends ZIOAppDefault {
  
  case class User(id: UUID, name: String) {
    def toJson = s"""{ "id": "$id", "name": "$name"}""".stripMargin
  }
  private val userLogAnnotation = LogAnnotation[User]("user", (_, i) => i, _.toJson)

  override val bootstrap: ZLayer[ZIOAppArgs, Any, Any] =
    Runtime.removeDefaultLoggers >>> consoleJson(
      LogFormat.default + LogFormat.annotation(LogAnnotation.TraceId) + LogFormat.annotation(
        userLogAnnotation
      )
    )

  private val users = List.fill(2)(UUID.randomUUID())

  override def run: ZIO[Scope, Any, ExitCode] =
    (for {
      traceId <- ZIO.succeed(UUID.randomUUID())
      _       <- ZIO.foreachPar(users) { uId =>
                   {
                     ZIO.logInfo("Starting operation") *>
                       ZIO.sleep(500.millis) *>
                       ZIO.logInfo("Stopping operation")
                   } @@ userLogAnnotation(User(uId, "aaa"))
                 } @@ LogAnnotation.TraceId(traceId)
      _       <- ZIO.logInfo("Done")
    } yield ExitCode.success)
}
```

however current json logger is not able recognise json string value

```json
{"timestamp":"2023-01-26T23:39:19.142999+01:00","level":"INFO","thread":"zio-fiber-5","message":"Starting operation","trace_id":"01087a2a-a46c-4ff5-b4de-1c3c680b55a6","user":"{ \"id\": \"d894df3a-265f-4fc2-8abf-1808e52b5c4f\", \"name\": \"aaa\"}"}
{"timestamp":"2023-01-26T23:39:19.143007+01:00","level":"INFO","thread":"zio-fiber-6","message":"Starting operation","trace_id":"01087a2a-a46c-4ff5-b4de-1c3c680b55a6","user":"{ \"id\": \"3e9a023e-9f4c-4ebb-aa9d-e231c8e17400\", \"name\": \"aaa\"}"}
```

this could be fixed, but it is probably not very safe and future proof

[...]

PR #657, opened 2023-02-14
Implement stringIsJson method

ipetkovic:
This is the WIP implementation of stringIsJson method.
Could you have a look and see if this makes sense please?

Some remarks:

- I am using custom small boolean stack to keep information about nested structure (current position inside array or inside an object). The requirement was to make this zero-allocation method which I believe is true here but I am using custom small sized boolean stack implemented with plain Long value to avoid heap allocation. Info I found on the net was contradictory on whether array can be allocated on stack or not. I tried few things but failed to make it stack allocated. In the end I decided for a Long value where bits represent state - 0 (inside array), 1 (inside object) which could be an overkill just to avoid heap allocation but I left it anyways :D The downside is that I cannot check strings which have nested depth larger than 58. One option would be to use preallocated array or perhaps not to track this info at all (then we could have false positives for JSON string)
- Number check is not strict - I am just checking for allowed characters there. Can easily change this as well. I was not sure how strict of a check we want

I am happy to hear any comments.

### DIFF
--- a/core/shared/src/main/scala/zio/logging/internal/JsonValidator.scala
+++ b/core/shared/src/main/scala/zio/logging/internal/JsonValidator.scala
@@ -1,0 +1,202 @@
+package zio.logging.internal
+
+import scala.annotation.tailrec
+
+object JsonValidator {
+
+  def isJson(string: String): Boolean = {
+    val startIdx = stringCouldBeJson(string)
+    if (startIdx == NOT_A_JSON) false
+    else
+      checkJson(
+        string,
+        startIdx,
+        STATE_EXPECTING_VALUE,
+        if (string(startIdx) == '{') OBJECT
+        else ARRAY
+      ) == string.length
+  }
+
+  private def checkJson(
+    string: String,
+    startIdx: Int,
+    startState: Byte,
+    blockType: Boolean
+  ): Int = {
+    var idx   = startIdx
+    var state = startState
+
+    while (state != STATE_BLOCK_END && idx < string.length) {
+      val c = string(idx)
+
+      if (isWhitespace(c)) idx += 1
+      else if (state == STATE_EXPECTING_VALUE) {
+
+        if (c == '{') idx = checkJson(string, idx + 1, STATE_BLOCK_START, OBJECT)
+        else if (c == '[') idx = checkJson(string, idx + 1, STATE_BLOCK_START, ARRAY)
+        else idx = skipSimpleValue(string, idx)
+
+        state = STATE_VALUE_PROCESSED
+
+      } else if (state == STATE_VALUE_PROCESSED)
+        if (c == ',' && blockType == OBJECT) {
+          idx = handleObjectKey(string, skipWhitespaces(string, idx + 1))
+          state = STATE_EXPECTING_VALUE
+        } else if (c == ',' && blockType == ARRAY) {
+          idx += 1
+          state = STATE_EXPECTING_VALUE
+        } else if ((c == '}' && blockType == OBJECT) || (c == ']' && blockType == ARRAY)) {
+          idx += 1
+          state = STATE_BLOCK_END
+        } else
+          idx = NOT_A_JSON
+      else if (state == STATE_BLOCK_START)
+        if (((c == '}' && blockType == OBJECT) || (c == ']' && blockType == ARRAY))) {
+          idx += 1
+          state = STATE_BLOCK_END
+        } else if (blockType == OBJECT) {
+          idx = handleObjectKey(string, skipWhitespaces(string, idx))
+          state = STATE_EXPECTING_VALUE
+        } else
+          state = STATE_EXPECTING_VALUE
+      else idx = NOT_A_JSON
+    }
+
+    idx
+  }
+
+  private def skipSimpleValue(string: String, startIdx: Int) =
+    if (startIdx < string.length) {
+      val c = string(startIdx)
+      if (c == '"')
+        skipString(string, startIdx)
+      else if (isDigit(c) || c == '-')
+        skipNumber(string, startIdx)
+      else if (c == 'f' || c == 'n' || c == 't')
+        skipBooleanOrNull(string, startIdx)
+      else NOT_A_JSON
+    } else NOT_A_JSON
+
+  private def handleObjectKey(string: String, startIdx: Int) = {
+    val idx = skipWhitespaces(string, skipString(string, startIdx))
+    if (idx < string.length && string(idx) == ':') idx + 1
+    else NOT_A_JSON
+  }
+
+  private def stringCouldBeJson(string: String): Int = {
+    val idxStart = skipWhitespaces(string, 0)
+    val idxEnd   = skipWhitespacesBackwards(string, string.length - 1)
+
+    if (idxStart >= string.length || idxEnd <= 0) NOT_A_JSON
+    else if (string.charAt(idxStart) == '{' & string.charAt(idxEnd) == '}') idxStart
+    else if (string.charAt(idxStart) == '[' & string.charAt(idxEnd) == ']') idxStart
+    else NOT_A_JSON
+  }
+
+  private def skipBooleanOrNull(string: String, idx: Int): Int =
+    if (idx + 3 >= string.length) NOT_A_JSON
+    else {
+      val c1       = string(idx)
+      val c2       = string(idx + 1)
+      val c3       = string(idx + 2)
+      val c4       = string(idx + 3)
+      val fifthIsE = (idx + 4 < string.length) && string(idx + 4) == 'e'
+
+      if (c1 == 't' && c2 == 'r' && c3 == 'u' && c4 == 'e')
+        idx + 4
+      else if (c1 == 'f' && c2 == 'a' && c3 == 'l' && c4 == 's' && fifthIsE)
+        idx + 5
+      else if (c1 == 'n' && c2 == 'u' && c3 == 'l' && c4 == 'l')
+        idx + 4
+      else
+        NOT_A_JSON
+    }
+
+  private def skipString(string: String, startIdx: Int): Int = {
+    @tailrec def inner(string: String, idx: Int, isBackslashed: Boolean): Int =
+      if (idx >= string.length) NOT_A_JSON
+      else if (!isBackslashed && string(idx) == '"') idx + 1
+      else inner(string, idx + 1, !isBackslashed && string(idx) == '\\')
+
+    if (startIdx < string.length && string(startIdx) == '"')
+      inner(string, startIdx + 1, false)
+    else NOT_A_JSON
+  }
+
+  @tailrec private def skipWhitespaces(string: String, idx: Int): Int =
+    if (idx >= string.length) NOT_A_JSON
+    else if (isWhitespace(string.charAt(idx))) skipWhitespaces(string, idx + 1)
+    else idx
+
+  @inline private def isWhitespace(c: Char) =
+    c <= ' ' && whitespaceLookup(c)
+
+  @inline private def isDigit(c: Char) = c >= '0' && c <= '9'
+
+  @tailrec private def skipWhitespacesBackwards(string: String, idx: Int): Int =
+    if (idx >= string.length || idx <= 0) idx
+    else if (!isWhitespace(string.charAt(idx))) idx
+    else skipWhitespacesBackwards(string, idx - 1)
+
+  @tailrec private def skipDigits(string: String, idx: Int): Int =
+    if (idx >= string.length) NOT_A_JSON
+    else if (!isDigit(string.charAt(idx))) idx
+    else skipDigits(string, idx + 1)
+
+  private def skipExponentPart(string: String, startIdx: Int) =
+    if (startIdx < string.length && (string(startIdx) == 'e' || string(startIdx) == 'E')) {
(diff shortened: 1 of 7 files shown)


core/shared/src/main/scala/zio/logging/internal/json/SmallBooleanStack.scala
```diff
@@ -0,0 +1,35 @@
...
+  @inline final private def setState(state: Boolean): Long =
+    if (state) data | (1L << stateIdx)
+    else data & ~(1L << stateIdx)
+
+  def isDefined()      = data != undefinedValue
+  def isEmpty: Boolean = stateIdx == 0
+
+  def pop(): SmallBooleanStack = {
+    if (stateIdx == 0) new SmallBooleanStack(undefinedValue)
+    new SmallBooleanStack(data - idxIncrement)
```
justcoon, line 23: should not be `else` here ?

core/shared/src/main/scala/zio/logging/internal/json/SmallBooleanStack.scala
```diff
@@ -0,0 +1,35 @@
+package zio.logging.internal.json
+
+final class SmallBooleanStack(
+  private val data: Long
+) extends AnyVal {
+  /* 64 - stackSize reserved for index store -> 6 bits are enough to represent 58 states */
+  @inline final private def stackSize      = 58
+  @inline final private def idxIncrement   = 0x400000000000000L // 1L << stackSize
+  @inline final private def undefinedValue = 0xffffffffffffffffL
```
justcoon, line 9: maybe those 3 defs should be in companion object as val

core/shared/src/main/scala/zio/logging/internal/json/SmallBooleanStack.scala
```diff
@@ -0,0 +1,35 @@
...
+  def isEmpty: Boolean = stateIdx == 0
+
+  def pop(): SmallBooleanStack = {
+    if (stateIdx == 0) new SmallBooleanStack(undefinedValue)
+    new SmallBooleanStack(data - idxIncrement)
+  }
+
+  def push(value: Boolean): SmallBooleanStack =
+    if (stateIdx >= stackSize - 1) new SmallBooleanStack(undefinedValue)
+    else new SmallBooleanStack(setState(value)).increaseIdx()
```
justcoon, line 28: this branch creating SmallBooleanStack twice (second time in increaseIdx)

core/shared/src/main/scala/zio/logging/internal/json/SmallBooleanStack.scala
```diff
@@ -0,0 +1,35 @@
...
+) extends AnyVal {
+  /* 64 - stackSize reserved for index store -> 6 bits are enough to represent 58 states */
+  @inline final private def stackSize      = 58
+  @inline final private def idxIncrement   = 0x400000000000000L // 1L << stackSize
+  @inline final private def undefinedValue = 0xffffffffffffffffL
+
+  @inline final private def stateIdx      = data >>> stackSize
+  @inline final private def increaseIdx() = new SmallBooleanStack(data + idxIncrement)
+
+  @inline final private def setState(state: Boolean): Long =
```
justcoon, line 14: setState is private, but name is confusing as it look like mutating function ...

core/shared/src/main/scala/zio/logging/internal/json/SmallBooleanStack.scala
```diff
@@ -0,0 +1,35 @@
...
+    if (state) data | (1L << stateIdx)
+    else data & ~(1L << stateIdx)
+
+  def isDefined()      = data != undefinedValue
+  def isEmpty: Boolean = stateIdx == 0
+
+  def pop(): SmallBooleanStack = {
+    if (stateIdx == 0) new SmallBooleanStack(undefinedValue)
```
justcoon, line 22: maybe 
```scala
object SmallBooleanStack {
    private val undefinedValue = 0xffffffffffffffffL
    
    val undefined: SmallBooleanStack = new SmallBooleanStack(undefinedValue)
}

```

core/shared/src/main/scala/zio/logging/internal/json/SmallBooleanStack.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  def push(value: Boolean): SmallBooleanStack =
+    if (stateIdx >= stackSize - 1) new SmallBooleanStack(undefinedValue)
+    else new SmallBooleanStack(setState(value)).increaseIdx()
+
+  def top(): Boolean = ((data >>> stateIdx - 1) & 1) == 1
+}
+
+object SmallBooleanStack {
+  def apply() = new SmallBooleanStack(0L)
```
justcoon, line 34: maybe 
``` scala
val empty: SmallBooleanStack = new SmallBooleanStack(0L)
```

core/shared/src/main/scala/zio/logging/internal/json/JsonValidator.scala
```diff
@@ -0,0 +1,165 @@
...
+          currentState = STATE_EXPECTING_VALUE
+        }
+      } else idx = NOT_A_JSON
+    }
+
+    idx != NOT_A_JSON && stack.isEmpty
+  }
+
+  private def skipSimpleValue(string: String, startIdx: Int) = {
+    val c = string(startIdx)
```
justcoon, line 82: add check 
```if (startIdx < string.length)```

**Comment by justcoon** [comment]
@ipetkovic thank you for contribution

I added some comments, maybe it is good to add some tests, 

we will need them anyway, also even if this impl. will change, tests should not a lot

**Comment by ipetkovic** [comment]
@justcoon Thanks for comments, will update it today probably, also with tests. I already have some but need to add some more cases.
Can you just check remarks in my original comments and let me know what you think?

**Comment by justcoon** [comment]
@ipetkovic 
personally I think, in this case it is about the ratio between performance and correctness,
if we can do very fast "partial" validation in compare to completely correct validation (but much slower one), i think it should be ok in this case

maybe it could be good to check (if you did not do it already) how some others json libraries doing it, or do "validation" benchmark - this implementation vs. some json library implementation,
i did not seen some latest benchmarks in relation to json java or scala libraries, but some time ago I seen that jsoniter was mentioned like fast one 

also considering that this will be internal impl. we can do some improvements in next releases ....

File: core/shared/src/main/scala/zio/logging/internal/json/JsonValidator.scala [hunk 8]
```diff
@@ -0,0 +1,165 @@
...
+    val startIdx = stringCouldBeJson(string)
+    if (startIdx == NOT_A_JSON) false
+    else checkJson(string, startIdx)
+  }
+
+  private def checkJson(
+    string: String,
+    startIdx: Int
+  ): Boolean = {
+    var stack        = SmallBooleanStack()
```
jdegoes, line 17: The trick to eliminate use of `SmallBooleanStack` is to _recurse_: this allows you to store on the JVM stack the information on the block type. "Popping" the stack corresponds to exiting one of the recursions, and bubbling up to the next higher-level; "pushing" a boolean on the stack corresponds to recursing.

In order to implement this optimization, you can generalize `checkJson` to take the current state and the boolean. 

Note that when completed, this means we cannot validate JSON of arbitrary depth. We could solve that by switching over to a slower-based validator after we recurse `n` times (where `n` > 200, say). But I'd say solving that is not essential for v1.

jdegoes:
@ipetkovic Overall, great work coming up with this small, efficient validator! I made one suggestion to eliminate the heap usage.

The other thing we need hear is a `Gen`-based test. We need to generate random JSON and verify that this will validate all JSON correctly, even weird JSON with non-ASCII characters, etc.

core/shared/src/main/scala/zio/logging/internal/json/JsonValidator.scala
ipetkovic: Done

core/shared/src/main/scala/zio/logging/internal/json/SmallBooleanStack.scala
ipetkovic: Not relevant anymore.

core/shared/src/main/scala/zio/logging/internal/json/SmallBooleanStack.scala
ipetkovic: Not relevant anymore.

core/shared/src/main/scala/zio/logging/internal/json/SmallBooleanStack.scala
ipetkovic: Not relevant anymore.

core/shared/src/main/scala/zio/logging/internal/json/SmallBooleanStack.scala
ipetkovic: Not relevant anymore.

core/shared/src/main/scala/zio/logging/internal/json/SmallBooleanStack.scala
ipetkovic: Not relevant anymore.

core/shared/src/main/scala/zio/logging/internal/json/SmallBooleanStack.scala
ipetkovic: Not relevant anymore.

core/shared/src/main/scala/zio/logging/internal/json/JsonValidator.scala
ipetkovic: Done.

ipetkovic:
Thank you both for the comments! I updated the implementation to be more efficient using recursive approach.
@jdegoes  Thanks for suggestion for recursion , seemed so obvious after you mentioned it :D
@justcoon it should be strict now. Also, more strict check for numbers makes it even more efficient than my last simple solution. It kind of makes sense since checking for all allowed characters in number with lots of digits is more expensive than checking for digit only for example etc.

Gen-based tests will come soon.

ipetkovic:
@justcoon @jdegoes I did some benchmarking and improved an algorithm a bit. I attached test for JsonValidator and included the fix in LogAppender.

Compared to my first version, this one handles whitespaces much better now. It was a bit naive to use isWhitespace method from RichChar. Now I introduced lookup array for whitespace which improved performance for up to factor of 2.
I noticed this issue with pretty printed JSON files.

I measured times to process 15 different JSON files (500 bytes for up to 700KB) and compared it with play and gson parsers. It is still not fair comparison since these parsers do much more but still...

- worst case: JsonValidator 30% faster
- best case: JsonValidator 185% faster
- average: JsonValidator 88% faster
- median: JsonValidator 91% faster

, play is pretty slow:

- worst case: JsonValidator 100% faster
- best case: JsonValidator 830% faster
- average : JsonValidator 275% faster

I tried to use jsoniter but failed to make it work for an arbitrary JSON structure. Did not spend too much time there but I can in the next days if you like.

Benchmark was done using -XX:+UseEpsilonGC, disabled GC, + all memory preallocated.

There is room to further optimize the algorithm but the biggest issue is, in my opinion, the fact that I am working with String, and using charAt to access characters. Using underlying byte array would make it much more efficient. But this would then result in memory allocation since, as far as I know, it is not possible to get access to internal array. At least not in non-kinky way.

Please let me know what you think.

core/shared/src/main/scala/zio/logging/internal/JsonValidator.scala
```diff
@@ -0,0 +1,202 @@
+package zio.logging.internal
+
+import scala.annotation.tailrec
+
+object JsonValidator {
+
+  def stringIsJson(string: String): Boolean = {
```
justcoon, line 7: maybe 

stringIsJson -> isJson

justcoon:
@ipetkovic 

can you add tests/spec to JsonLogFormatSpec, which covering this feature?

core/shared/src/main/scala/zio/logging/internal/JsonValidator.scala
ipetkovic: Done

ipetkovic:
@justcoon I updated with one test. Do you think I should add more? Other tests are covering non-json part so I figured one should do it.

justcoon:
@ipetkovic , i recommend to add some spec for annotations with json value

also it will be good to update `ConsoleJsonApp` in  examples/core  (to use annotation with json value like it is in #643) or add another app, which show that json structured logging is working 
and then also update docs with that example 

thank you

ipetkovic:
@justcoon I updated JsonLogFormatSpec with few more test cases.
Example and docu are also updated.

justcoon (approved):

justcoon:
@ipetkovic - thank you for contribution

ipetkovic:
@justcoon No problem! It was fun :D
Can you please approve it once more? I squashed everything into one commit.

justcoon (approved):

ipetkovic:
@justcoon can you merge it? it seems I don't have write access

jdegoes (approved):

jdegoes:
@ipetkovic Thanks for your work on this! Great job making it 0 allocation. 💪

PR #664, opened 2023-02-21
Apply ZIO SBT Ecosystem Plugin

khajavi:


.github/workflows/ci.yml
```diff
@@ -1,110 +1,205 @@
...
-      - name: Mima Checks
-        if: ${{ !startsWith(matrix.scala, '3.') && endsWith(matrix.java, '.11') }}
-        run: ./sbt ++${{ matrix.scala }}! mimaChecks
-      - name: Run tests
-        if: ${{ endsWith(matrix.java, '.11') }}
-        run: sbt ++${{ matrix.scala }}! test${{ matrix.platform }}
-      - name: Run tests JVM8
-        if: ${{ endsWith(matrix.java, '.8') && matrix.platform == 'JVM' }}
-        run: sbt ++${{ matrix.scala }}! testJVM8
-      - name: Compile additional subprojects
```
justcoon, line 77: it is look like that this is missing in new version, if I see properly ....

khajavi: There is `+publishLocal` on line 39, but it doesn't compile against different java versions. If it is insufficient, I will add CI customization to the `zio-sbt-ci` plugin.

khajavi: @justcoon The question is when we are testing main components against various java versions, why do we need to test the compilation phase of their example projects?

justcoon: to check if they are compilable , as examples are in custom modules

khajavi: additional tests were incorporated.

justcoon (approved):

PR #667, opened 2023-02-25
fix scala 3 version checks in build

tellnobody1:
It is strange that it is working without a fix, so probably these cases are not needed at all.


justcoon:
@tellnobody1 thank you for contribution

PR #668, opened 2023-02-26
Logger config

justcoon:
loggers configuration support

added `LogPattern` , which is string representation of `LogFormat` - zio-parser is used for parsing

removed scala 2.11 support


jdegoes:
1. Yes, I'd drop Scala 2.11.
2. A class called `X` should contain constructors in companion object of `X`. The pattern can be broken if necessary, but it is always confusing to do so.

core/shared/src/main/scala/zio/logging/LogPattern.scala
```diff
@@ -0,0 +1,368 @@
...
+import zio.parser.{ Syntax, _ }
+import zio.{ Chunk, Config }
+
+import java.time.format.DateTimeFormatter
+import scala.util.Try
+
+/**
+ * A `LogPattern` is string representation of `LogFormat`
+ */
+sealed trait LogPattern {
```
jdegoes, line 27: I wonder if we should call this `SerializableLogFormat`, or perhaps, `LogFormat.Serializable` or something.

The presence of `toLogFormat` suggests that it might be or should be a subtype of `LogFormat`.

Expressing that relationship in the name could be useful. Also moving it to the companion object of `LogFormat` could be useful if it's not as commonly used as `LogFormat`, and we want to hide it a bit or emphasize it's a derivative of `LogFormat`.

core/shared/src/main/scala/zio/logging/LogPattern.scala
```diff
@@ -0,0 +1,368 @@
...
+  def isDefinedFilter: Option[LogFilter[Any]] = None
+}
+
+object LogPattern {
+
+  sealed trait Arg extends LogPattern {
+    def name: String
+  }
+
+  sealed trait Arg1[A1] extends Arg {
```
jdegoes, line 43: The type parameters here become erased because they are not tracked in the supertype.

Which means this is practically equivalent to `def arg1: Any`, etc. 

I wonder if, instead of doing it this way, we can either delete `Arg1` and `Arg2`, or embed type-safe capabilities into these treats, which lets clients of the traits use the arguments in some safe way?

jdegoes (dismissed):
Beautiful

jdegoes (approved):

PR #691, opened 2023-03-16
rolling log

SHSongs:
I use Slf4jBridge in my project, but zio-logging does not include rolling log function,
Therefore, I have created a simple rolling log function.
could you review the PR?

### DIFF
--- a/core/shared/src/main/scala/zio/logging/FileLoggerConfig.scala
+++ b/core/shared/src/main/scala/zio/logging/FileLoggerConfig.scala
@@ -28,7 +28,8 @@ final case class FileLoggerConfig(
   filter: LogFilter[String],
   charset: Charset = StandardCharsets.UTF_8,
   autoFlushBatchSize: Int = 1,
-  bufferedIOSize: Option[Int] = None
+  bufferedIOSize: Option[Int] = None,
+  rolling: Boolean
 )
 
 object FileLoggerConfig {
@@ -64,7 +65,8 @@ object FileLoggerConfig {
           LogFilter.logLevelByName(filterConfig),
           charset,
           autoFlushBatchSize,
-          bufferedIOSize
+          bufferedIOSize,
+          rolling = false
         )
     }
   }
--- a/core/shared/src/main/scala/zio/logging/internal/FileWriter.scala
+++ b/core/shared/src/main/scala/zio/logging/internal/FileWriter.scala
@@ -15,15 +15,18 @@
  */
 package zio.logging.internal
 
-import java.io.{ BufferedWriter, FileOutputStream, OutputStreamWriter, Writer }
+import java.io.{ BufferedWriter, File, FileOutputStream, OutputStreamWriter, Writer }
 import java.nio.charset.Charset
-import java.nio.file.Path
+import java.nio.file.{ FileSystems, Path }
+import java.time.LocalDateTime
+import java.time.format.DateTimeFormatter
 
 private[logging] class FileWriter(
   destination: Path,
   charset: Charset,
   autoFlushBatchSize: Int,
-  bufferedIOSize: Option[Int]
+  bufferedIOSize: Option[Int],
+  rolling: Boolean
 ) extends Writer {
   private val writer: Writer = {
     val output = new OutputStreamWriter(new FileOutputStream(destination.toFile, true), charset)
@@ -38,15 +41,46 @@ private[logging] class FileWriter(
   final def write(buffer: Array[Char], offset: Int, length: Int): Unit =
     writer.write(buffer, offset, length)
 
-  final def writeln(line: String): Unit = {
-    writer.write(line)
-    writer.write(System.lineSeparator)
-
-    entriesWritten += 1
-
-    if (entriesWritten % autoFlushBatchSize == 0)
-      writer.flush()
+  private def makeDateFile(): File = {
+    val currentDateTime = LocalDateTime.now()
+    val formatter       = DateTimeFormatter.ofPattern("yyyy-MM-dd")
+    val time            = formatter.format(currentDateTime)
+    val fileNameArray   = destination.getFileName.toString.split("\\.")
+    val timeFileName    = if (fileNameArray.length >= 2) {
+      fileNameArray.drop(1).mkString(".") + "-" + time + "." + fileNameArray.last
+    } else {
+      fileNameArray.head + "-" + time
+    }
+    val timeDestination = FileSystems.getDefault.getPath(destination.getParent.toString, timeFileName)
+    timeDestination.toFile
   }
+
+  final def writeln(line: String): Unit =
+    if (rolling) {
+      val output = new OutputStreamWriter(
+        new FileOutputStream(makeDateFile(), true),
+        charset
+      )
+      val writer = bufferedIOSize match {
+        case Some(bufferSize) => new BufferedWriter(output, bufferSize)
+        case None             => output
+      }
+      writer.write(line)
+      writer.write(System.lineSeparator)
+
+      entriesWritten += 1
+
+      if (entriesWritten % autoFlushBatchSize == 0)
+        writer.flush()
+    } else {
+      writer.write(line)
+      writer.write(System.lineSeparator)
+
+      entriesWritten += 1
+
+      if (entriesWritten % autoFlushBatchSize == 0)
+        writer.flush()
+    }
 
   final def flush(): Unit = writer.flush()
 
--- a/core/shared/src/main/scala/zio/logging/package.scala
+++ b/core/shared/src/main/scala/zio/logging/package.scala
@@ -169,9 +169,10 @@ package object logging {
     logLevel: LogLevel = LogLevel.Info,
     charset: Charset = StandardCharsets.UTF_8,
     autoFlushBatchSize: Int = 1,
-    bufferedIOSize: Option[Int] = None
+    bufferedIOSize: Option[Int] = None,
+    rolling: Boolean = false
   ): ZLayer[Any, Nothing, Unit] =
-    file(destination, format, LogFilter.logLevel(logLevel), charset, autoFlushBatchSize, bufferedIOSize)
+    file(destination, format, LogFilter.logLevel(logLevel), charset, autoFlushBatchSize, bufferedIOSize, rolling)
 
   @deprecated("use zio.logging.fileLogger", "2.1.10")
   def file(
@@ -180,9 +181,10 @@ package object logging {
     logFilter: LogFilter[String],
     charset: Charset,
     autoFlushBatchSize: Int,
-    bufferedIOSize: Option[Int]
+    bufferedIOSize: Option[Int],
+    rolling: Boolean
   ): ZLayer[Any, Nothing, Unit] =
-    fileLogger(FileLoggerConfig(destination, format, logFilter, charset, autoFlushBatchSize, bufferedIOSize))
+    fileLogger(FileLoggerConfig(destination, format, logFilter, charset, autoFlushBatchSize, bufferedIOSize, rolling))
 
   @deprecated("use zio.logging.fileAsyncLogger", "2.1.10")
   def fileAsync(
@@ -191,7 +193,8 @@ package object logging {
     logLevel: LogLevel = LogLevel.Info,
     charset: Charset = StandardCharsets.UTF_8,
     autoFlushBatchSize: Int = 1,
-    bufferedIOSize: Option[Int] = None
+    bufferedIOSize: Option[Int] = None,
+    rolling: Boolean
   ): ZLayer[Any, Nothing, Unit] =
     fileAsync(
       destination,
@@ -199,7 +202,8 @@ package object logging {
       LogFilter.logLevel(logLevel),
       charset,
       autoFlushBatchSize,
-      bufferedIOSize
+      bufferedIOSize,
+      rolling
     )
 
   @deprecated("use zio.logging.fileAsyncLogger", "2.1.10")
@@ -209,9 +213,12 @@ package object logging {
     logFilter: LogFilter[String],
     charset: Charset,
     autoFlushBatchSize: Int,
-    bufferedIOSize: Option[Int]
+    bufferedIOSize: Option[Int],
+    rolling: Boolean
   ): ZLayer[Any, Nothing, Unit] =
-    fileAsyncLogger(FileLoggerConfig(destination, format, logFilter, charset, autoFlushBatchSize, bufferedIOSize))
+    fileAsyncLogger(
+      FileLoggerConfig(destination, format, logFilter, charset, autoFlushBatchSize, bufferedIOSize, rolling)
+    )
 
   @deprecated("use zio.logging.fileJsonLogger", "2.1.10")
   def fileJson(
@@ -220,9 +227,10 @@ package object logging {
     logLevel: LogLevel = LogLevel.Info,
     charset: Charset = StandardCharsets.UTF_8,
     autoFlushBatchSize: Int = 1,
-    bufferedIOSize: Option[Int] = None
+    bufferedIOSize: Option[Int] = None,
+    rolling: Boolean
   ): ZLayer[Any, Nothing, Unit] =
-    fileJson(destination, format, LogFilter.logLevel(logLevel), charset, autoFlushBatchSize, bufferedIOSize)
+    fileJson(destination, format, LogFilter.logLevel(logLevel), charset, autoFlushBatchSize, bufferedIOSize, rolling)
 
   @deprecated("use zio.logging.fileJsonLogger", "2.1.10")
   def fileJson(
@@ -231,10 +239,11 @@ package object logging {
     logFilter: LogFilter[String],
     charset: Charset,
     autoFlushBatchSize: Int,
-    bufferedIOSize: Option[Int]
+    bufferedIOSize: Option[Int],
+    rolling: Boolean
   ): ZLayer[Any, Nothing, Unit] =
     fileJsonLogger(
-      FileLoggerConfig(destination, format, logFilter, charset, autoFlushBatchSize, bufferedIOSize)
+      FileLoggerConfig(destination, format, logFilter, charset, autoFlushBatchSize, bufferedIOSize, rolling)
     )
 
   @deprecated("use zio.logging.fileAsyncJsonLogger", "2.1.10")
@@ -244,7 +253,8 @@ package object logging {
     logLevel: LogLevel = LogLevel.Info,
     charset: Charset = StandardCharsets.UTF_8,
     autoFlushBatchSize: Int = 1,
-    bufferedIOSize: Option[Int] = None
+    bufferedIOSize: Option[Int] = None,
+    rolling: Boolean
   ): ZLayer[Any, Nothing, Unit] =
     fileAsyncJson(
       destination,
@@ -252,7 +262,8 @@ package object logging {
       LogFilter.logLevel(logLevel),
       charset,
       autoFlushBatchSize,
-      bufferedIOSize
+      bufferedIOSize,
+      rolling
     )
 
   @deprecated("use zio.logging.fileAsyncJsonLogger", "2.1.10")
@@ -262,10 +273,11 @@ package object logging {
     logFilter: LogFilter[String],
     charset: Charset,
     autoFlushBatchSize: Int,
-    bufferedIOSize: Option[Int]
+    bufferedIOSize: Option[Int],
+    rolling: Boolean
   ): ZLayer[Any, Nothing, Unit] =
     fileAsyncJsonLogger(
-      FileLoggerConfig(destination, format, logFilter, charset, autoFlushBatchSize, bufferedIOSize)
+      FileLoggerConfig(destination, format, logFilter, charset, autoFlushBatchSize, bufferedIOSize, rolling)
     )
 
   def fileAsyncJsonLogger(config: FileLoggerConfig): ZLayer[Any, Nothing, Unit] =
@@ -354,7 +366,8 @@ package object logging {
     config.filter,
     config.charset,
     config.autoFlushBatchSize,
-    config.bufferedIOSize
+    config.bufferedIOSize,
+    config.rolling
   )
 
   private def makeFileAsyncLogger(
@@ -365,7 +378,8 @@ package object logging {
     config.filter,
     config.charset,
     config.autoFlushBatchSize,
-    config.bufferedIOSize
+    config.bufferedIOSize,
+    config.rolling
   )
 
   private def makeFileAsyncLogger(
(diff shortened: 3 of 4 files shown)


justcoon:
hi @SHSongs 

thank you for contribution, that is missing feature which will be nice to add,  I appreciate yours contribution

however there are few things to improve:

1. instead of simple `FileLoggerConfig.rolling` boolean flag, we need to be more open, if you take https://logback.qos.ch/manual/appenders.html#onRollingPolicies as inspiration, 
we will need to add something like 

```scala
sealed trait FileRollingPolicy

object FileRollingPolicy {
   case class TimeBasedRollingPolicy( /* some attributes related to this policy if needed */) extends FileRollingPolicy 
   // case class FixedWindowRollingPolicy() extends FileRollingPolicy   // can be implemented later ...
}

```
then
 
```scala
final case class FileLoggerConfig(
  ...
  bufferedIOSize: Option[Int] = None
  rollingPolicy: Option[FileRollingPolicy] = None
)
```
then this needs to be added to FileLoggerConfig.config ...

2. FileWriter need to work with this policy, if there is new file based on policy OutputStreamWriter needs to be opened just once, also you need to close previous one if there is new one created based on changes in relation to rolling policy

core/shared/src/main/scala/zio/logging/package.scala
```diff
@@ -220,9 +227,10 @@ package object logging {
     logLevel: LogLevel = LogLevel.Info,
     charset: Charset = StandardCharsets.UTF_8,
     autoFlushBatchSize: Int = 1,
-    bufferedIOSize: Option[Int] = None
+    bufferedIOSize: Option[Int] = None,
+    rolling: Option[FileLoggerConfig.FileRollingPolicy]
```
justcoon, line 231:
```
rolling: Option[FileLoggerConfig.FileRollingPolicy] = None
```
as previous attributes have default value

core/shared/src/main/scala/zio/logging/internal/FileWriter.scala
```diff
@@ -38,15 +44,50 @@ private[logging] class FileWriter(
...
-    entriesWritten += 1
+  final def writeln(line: String): Unit =
+    rolling match {
+      case Some(value) =>
+        value match {
+          case FileRollingPolicy.TimeBasedRollingPolicy =>
+            val output = new OutputStreamWriter(
+              new FileOutputStream(makeDateFile(), true),
+              charset
+            )
```
justcoon, line 69: this in general need some major improvement (as I mentioned in my previous comment)
open output stream with each log message (writeln) is expensive  operation ...

also when FileWriter have rolling policy it does not make sense to create non rolling writer (line 34)

core/shared/src/main/scala/zio/logging/FileLoggerConfig.scala
```diff
@@ -28,10 +28,25 @@ final case class FileLoggerConfig(
...
+  sealed trait FileRollingPolicy
+  object FileRollingPolicy {
+    case object TimeBasedRollingPolicy extends FileRollingPolicy
+
+    private[logging] val logLevelMapping: Map[String, FileRollingPolicy] = Map(
+      "TimeBasedRollingPolicy" -> FileRollingPolicy.TimeBasedRollingPolicy
+    )
+
+    private[logging] def fileRollingPolicyValue(value: String): Either[Config.Error.InvalidData, FileRollingPolicy] =
+      logLevelMapping.get(value.toUpperCase) match {
```
justcoon, line 45: `value.toUpperCase` will not match with "TimeBasedRollingPolicy"

SHSongs:
I think I fixed all the code you reviewed. Thanks for your time and effort reviewed my code.

core/shared/src/main/scala/zio/logging/internal/FileWriter.scala
```diff
@@ -15,30 +15,75 @@
...
+  }
+
   final def writeln(line: String): Unit = {
+    rolling match {
+      case Some(policy) =>
+        policy match {
+          case FileRollingPolicy.TimeBasedRollingPolicy =>
+            val newPath = makeDatePath()
+            if (newPath != currentDestination) {
+              println("새로운 파일 생성")
```
justcoon, line 79: i assume this is for "Debug" purposes and it will be removed ...

core/shared/src/main/scala/zio/logging/internal/FileWriter.scala
```diff
@@ -15,30 +15,75 @@
...
-    val output = new OutputStreamWriter(new FileOutputStream(destination.toFile, true), charset)
+  private var currentDestination             = destination
+  private def makeWriter(path: Path): Writer = {
+    val output = new OutputStreamWriter(new FileOutputStream(path.toFile, true), charset)
     bufferedIOSize match {
       case Some(bufferSize) => new BufferedWriter(output, bufferSize)
       case None             => output
     }
   }
+  private var writer: Writer                 = rolling match {
```
justcoon, line 42: in case of rollingPolicy, i think some concurrent constructs will have to be used as at same time 2 threads/fibers may try to log, and we do not want to create new writer twice ...

core/shared/src/main/scala/zio/logging/internal/FileWriter.scala
```diff
@@ -15,30 +15,75 @@
...
+    timeDestination
+  }
+
   final def writeln(line: String): Unit = {
+    rolling match {
+      case Some(policy) =>
+        policy match {
+          case FileRollingPolicy.TimeBasedRollingPolicy =>
+            val newPath = makeDatePath()
+            if (newPath != currentDestination) {
```
justcoon, line 78: we will need more optimal comparison probably just based  on `currentDateTime`, 
as `makeDatePath` is relatively complex, to do it with every log line

we need to have logging as fast as possible, we need to avoid unnecessary memory allocations and computations

core/shared/src/main/scala/zio/logging/FileLoggerConfig.scala
```diff
@@ -28,10 +28,25 @@ final case class FileLoggerConfig(
...
+  bufferedIOSize: Option[Int] = None,
+  rollingPolicy: Option[FileLoggerConfig.FileRollingPolicy] = None
 )
 
 object FileLoggerConfig {
+  sealed trait FileRollingPolicy
+  object FileRollingPolicy {
+    case object TimeBasedRollingPolicy extends FileRollingPolicy
+
+    private[logging] val logLevelMapping: Map[String, FileRollingPolicy] = Map(
```
justcoon, line 40: logLevelMapping -> fileRollingPolicyMapping

core/shared/src/main/scala/zio/logging/FileLoggerConfig.scala
```diff
@@ -28,10 +28,25 @@ final case class FileLoggerConfig(
...
+      "TimeBasedRollingPolicy" -> FileRollingPolicy.TimeBasedRollingPolicy
+    )
+
+    private[logging] def fileRollingPolicyValue(value: String): Either[Config.Error.InvalidData, FileRollingPolicy] =
+      logLevelMapping.get(value) match {
+        case Some(v) => Right(v)
+        case None    => Left(Config.Error.InvalidData(Chunk.empty, s"Expected a LogLevel, but found ${value}"))
```
justcoon, line 47: Expected a LogLevel -> Expected a FileRollingPolicy

core/shared/src/main/scala/zio/logging/internal/FileWriter.scala
SHSongs: I rewrote. Could you check this commit? https://github.com/zio/zio-logging/pull/691/commits/15d17687e4a2edcbc63c678d75e225333899b433

core/shared/src/main/scala/zio/logging/internal/FileWriter.scala
```diff
@@ -48,7 +56,68 @@ private[logging] class FileWriter(
       writer.flush()
   }
 
-  final def flush(): Unit = writer.flush()
+  final def flush(): Unit = writerProvider.writer.flush()
+
+  final def close(): Unit = writerProvider.writer.close()
+
+  private[logging] sealed trait WriterProvider {
```
justcoon, line 63: now `WriterProvider` is inside `FileWriter`

i think you can move it out, so also it can be tested separately

justcoon:
also documentation will need to be updated https://github.com/zio/zio-logging/blob/master/docs/file-logger.md

core/shared/src/main/scala/zio/logging/internal/FileWriter.scala
```diff
@@ -84,18 +85,17 @@ private[logging] class FileWriter(
...
-
-      private def makeDatePath: Path = {
-        val currentDateTime = LocalDateTime.now()
-        val formatter       = DateTimeFormatter.ofPattern("yyyy-MM-dd-HH-mm:ss")
-        val time            = formatter.format(currentDateTime)
+      private var timeInUse = makeNewTime
+      private var currentWriter: Writer = makeWriter(makePath(timeInUse))
+      private def makeNewTime = LocalDateTime.now().truncatedTo(ChronoUnit.DAYS)
+      private def makePath(time: LocalDateTime): Path = {
+        val formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd")
```
justcoon, line 92: DateTimeFormatter should be created just once

SHSongs:
https://github.com/zio/zio-logging/blob/742bcbf0c651687a3d88152777a89a8bc8c3fdf8/core/shared/src/main/scala/zio/logging/LogFormat.scala#L798

i want to TestClock but not work. I think this part is the cause

so I used this  `loggint~~.repeatWhileZIO(_ => ZIO.succeed(true).delay(100.millisecond))` in testing.

if this PR is close. Could I change java time to ZIO time?

justcoon:
@SHSongs 

`zio.ZLogger` in is not effectful

that is one of reason why in those underlying impl. of loggers we are not using zio.* (Console, Clock ...) if it is not really needed

you can create custom format using `zio.Unsafe` to do some testing

```scala
  def timestamp(formatter: => DateTimeFormatter): LogFormat =
    text {
      val d = zio.Unsafe.unsafe { implicit u =>
        Runtime.default.unsafe.run(zio.Clock.instant).getOrThrow()
      }
      formatter.format(d.atZone(ZoneId.systemDefault()))
    }
```

but i think in `LogFormat.timestamp` implementation we do not want to change it now, as with `zio.Unsafe` it is more expensive operation then `ZonedDateTime.now()`, and logging should have minimum impact on performance of system

core/shared/src/main/scala/zio/logging/internal/FileWriter.scala
SHSongs: I wroted, Could you check this commit? https://github.com/zio/zio-logging/pull/691/commits/5dd2e73bdadd2d54abb16e9c9cf0296f4b71ba6d

SHSongs:
> also documentation will need to be updated https://github.com/zio/zio-logging/blob/master/docs/file-logger.md

I updated the documentation. let me know if there is any missing information or anything that needs to be added. 
https://github.com/zio/zio-logging/commit/90edf388177b088209bf6038f0328e5823f6da6f

docs/file-logger.md
```diff
@@ -101,6 +105,8 @@ object FileApp extends ZIOAppDefault {
 }
 ```
 
+Expected file name: `file_app-2023-03-05.log`
+
```
justcoon, line 109: Expected file content
bellow by timestamps do not match with file name

core/jvm/src/test/scala/zio/logging/internal/WriterProviderSpec.scala
```diff
@@ -0,0 +1,20 @@
...
+
+import java.nio.file.FileSystems
+import java.time.LocalDate
+import java.time.format.DateTimeFormatter
+
+object WriterProviderSpec extends ZIOSpecDefault {
+
+  val spec: Spec[Environment, Any] = suite("WriterProvider")(
+    test("Make path include date") {
+      val destination = FileSystems.getDefault.getPath("/tmp/file_app.log")
```
justcoon, line 13: please add also test for paths like:

- /tmp/file_app
- /tmp/file.app.log
- /tmp/file.app.out.log

core/shared/src/main/scala/zio/logging/internal/WriterProvider.scala
```diff
@@ -0,0 +1,73 @@
...
+
+    private var timeInUse             = makeNewTime
+    private var currentWriter: Writer = makeWriter(makePath(destination, timeInUse, formatter), charset, bufferedIOSize)
+
+    override def writer: Writer = {
+      val newTime = makeNewTime
+      if (newTime != timeInUse) {
+        currentWriter.close()
+        currentWriter = makeWriter(makePath(destination, timeInUse, formatter), charset, bufferedIOSize)
+        timeInUse = newTime
```
justcoon, line 45: as i mentioned before, i think this implementation is not thread safe, some concurrency patterns should be use so only one caller will create new writer and close current one if it is needed

docs/file-logger.md
```diff
@@ -84,6 +87,7 @@ object FileApp extends ZIOAppDefault {
        |logger {
        |  format = "%timestamp{yyyy-MM-dd'T'HH:mm:ssZ} %fixed{7}{%level} [%fiberId] %name:%line %message %cause"
        |  path = "file:///tmp/file_app.log"
+       |  rollingPolicy = TimeBasedRollingPolicy
```
justcoon, line 90: you should modify also https://github.com/zio/zio-logging/blob/master/examples/core/src/main/scala/zio/logging/example/FileApp.scala, to match with this documentation

core/shared/src/main/scala/zio/logging/internal/WriterProvider.scala
```diff
@@ -0,0 +1,73 @@
...
+      }
+      currentWriter
+    }
+  }
+  object TimeBasedRollingWriterProvider {
```
justcoon, line 50: maybe you can add 
```scala
val dateTimeFormatter: DateTimeFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd")
```
here to companion object, so it will be crated just once, and you don't have to have it like argument in `makePath`

core/jvm/src/test/scala/zio/logging/internal/WriterProviderSpec.scala
SHSongs: I added test case

docs/file-logger.md
SHSongs: I applied

core/shared/src/main/scala/zio/logging/internal/WriterProvider.scala
SHSongs: Do I make it like the code below? Also, could you provide an example code for concurrent patterns?

```scala
    import java.util.concurrent.locks.ReentrantLock
    private val lock: ReentrantLock = new ReentrantLock()

    override def writer: Writer = {
      val newTime = makeNewTime
      if (newTime != timeInUse) {
        lock.lock()
        try {
          currentWriter.close()
          currentWriter = makeWriter(makePath(destination, timeInUse), charset, bufferedIOSize)
          timeInUse = newTime
        } finally {
          lock.unlock()
        }
      }
      currentWriter
    }
```

core/shared/src/main/scala/zio/logging/internal/WriterProvider.scala
```diff
@@ -0,0 +1,75 @@
...
+  final case class TimeBasedRollingWriterProvider(
+    destination: Path,
+    charset: Charset,
+    bufferedIOSize: Option[Int]
+  ) extends WriterProvider {
+    import TimeBasedRollingWriterProvider._
+
+    private var timeInUse             = makeNewTime
+    private var currentWriter: Writer = makeWriter(makePath(destination, timeInUse), charset, bufferedIOSize)
+    val formatter: DateTimeFormatter  = DateTimeFormatter.ofPattern("yyyy-MM-dd")
```
justcoon, line 38: is this used ?

core/shared/src/main/scala/zio/logging/internal/WriterProvider.scala
justcoon: this could work 

however i think time should be checked also under the lock

```scala
  override def writer: Writer = {
    val newTime = makeNewTime
    if (newTime != timeInUse) {
      lock.lock()
      try {
         if (newTime != timeInUse) {
            currentWriter.close() 
            currentWriter = makeWriter(makePath(destination, newTime), charset, bufferedIOSize) 
            timeInUse = newTime
        }
      } finally {
        lock.unlock()
      }
    } 
     currentWriter
  }
```
or just

```scala
 override def writer: Writer = {
    val newTime = makeNewTime
    
      lock.lock()
      try {
         if (newTime != timeInUse) {
            currentWriter.close() 
            currentWriter = makeWriter(makePath(destination, newTime), charset, bufferedIOSize) 
            timeInUse = newTime
        }
      } finally {
        lock.unlock()
      }
    
     currentWriter
  }
  ```
  
  i think that first version should be less locking, even if it is probably less correct

File: core/shared/src/main/scala/zio/logging/internal/WriterProvider.scala (hunk 16 above)
Comment (SHSongs) [reply core/shared/src/main/scala/zio/logging/internal/WriterProvider.scala]: Not use, I removed formatter in provider class

File: core/shared/src/main/scala/zio/logging/internal/WriterProvider.scala [hunk 17]
```diff
@@ -0,0 +1,83 @@
...
+            currentWriter.close()
+            currentWriter = makeWriter(makePath(destination, newTimeInLock), charset, bufferedIOSize)
+            timeInUse = newTimeInLock
+          }
+        } finally {
+          lock.unlock()
+        }
+      }
+      currentWriter
+    }
```
SHSongs, line 57: https://github.com/zio/zio-logging/pull/691#discussion_r1143799344 continuously..

I applied a concurrent pattern and checked the time under the lock.  
After acquiring the lock, I created a new timestamp value using `makeNewTime()`,
. because I think that if I were to use the time that was determined before acquiring the lock to create a new file, repeatedly create and close new writers.

thread
```
A: 
set(newTime) -> lock -> if(newTime..) -> create 
B: 
set(newTime) -> lock -> wait                 -> if(newTime..) -> create
```

justcoon:
``` scala 
   override def writer: Writer = {
      val newTime = makeNewTime
      if (newTime != timeInUse) { // if they are same it will not go inside if, so it will not be locked
        lock.lock()
        try {

// considering that time is date, basically it should not happen that newTimeInLock != newTime
// otherwise it will have to wait for lock more almost one day
          val newTimeInLock = makeNewTime 
        
         // considering proposed    if (newTime != timeInUse) this additional check here in lock,
         // this is for situation, when 2 threads at same time try to obtain lock, (after first if condition)
        // one will get lock first, and this will change writer and timeInUse
        // then after lock will be released, second one will get lock, and enter here
        // but timeInUse and writer are already changed, 
        // so it should not be needed to recreate basically same thing
          if (newTimeInLock != timeInUse) {
            currentWriter.close()
            currentWriter = makeWriter(makePath(destination, newTimeInLock), charset, bufferedIOSize)
            timeInUse = newTimeInLock
          }
        } finally {
          lock.unlock()
        }
      }
      currentWriter
    }
```

SHSongs: I understand the comments (timeInUse and writer are already changed), and I rewrote code.

core/shared/src/main/scala/zio/logging/internal/WriterProvider.scala
```diff
@@ -0,0 +1,81 @@
...
+        case Some(bufferSize) => new BufferedWriter(output, bufferSize)
+        case None             => output
+      }
+    }
+  }
+
+  final case class TimeBasedRollingWriterProvider(
+    destination: Path,
+    charset: Charset,
+    bufferedIOSize: Option[Int]
```
justcoon, line 32: it will be good if we can test it

if we add here another attribute

```scala 
makeNewTime: () => LocalDateTime = TimeBasedRollingWriterProvider.makeNewTime
```

then in WriterProviderSpec

it should be possible to do something like:

```scala
  val timeRef = new AtomicReference[LocalDateTime](LocalDateTime.now())
  
   timeRef.set(LocalDateTime.now().plusDays(1))
  
   val testMakeNewTime: () => LocalDateTime = () => timeRef.get()
```
and then test it if in case of parallel execution, when  `WriterProvider.writer` is called multiple times with same time, if it return same writer

SHSongs: I wrote test code. Could you check this commit? https://github.com/zio/zio-logging/pull/691/commits/4fe78ca9e9229d06e767afb044df6ae36fe4b434

core/jvm/src/test/scala/zio/logging/internal/WriterProviderSpec.scala
```diff
@@ -0,0 +1,54 @@
...
+      val destination4   = FileSystems.getDefault.getPath("/tmp/file.app.out.log")
+
+      assertTrue(
+        (makePath(destination1, localDateTime).toString == "/tmp/file_app-2023-03-21") &&
+          (makePath(destination2, localDateTime).toString == "/tmp/file_app-2023-03-21.log") &&
+          (makePath(destination3, localDateTime).toString == "/tmp/file.app-2023-03-21.log") &&
+          (makePath(destination4, localDateTime).toString == "/tmp/file.app.out-2023-03-21.log")
+      )
+    },
+    test("aa") {
```
justcoon, line 30: please add better description of test case

core/jvm/src/test/scala/zio/logging/internal/WriterProviderSpec.scala
```diff
@@ -0,0 +1,54 @@
...
+        destination = FileSystems.getDefault.getPath("/tmp/file_app"),
+        charset = StandardCharsets.UTF_8,
+        bufferedIOSize = Some(1),
+        makeNewTime = testMakeNewTime
+      )
+
+      val parallelExecution = ZIO
+        .foreachPar(1 to 5) { _ =>
+          ZIO.succeed(writerProvider.writer)
+        } @@ ZIOAspect.parallel(5)
```
justcoon, line 45: ZIOAspect.parallel(5) should not be needed

SHSongs:
I rewrote code. Could you check this commit?  https://github.com/zio/zio-logging/pull/691/commits/0aa8499dc9564d52b62883817a28bc6de29cebd0

justcoon:
@SHSongs  please run `fmt` and `fix`, CI is failing

justcoon:
@SHSongs did you accepted licence/cla ?

SHSongs:
Screen Shot
@justcoon 
Yes, I accepted license/cla. this is my Screen

PR #718, opened 2023-05-11
Slf4j bridge - fiber propagation

justcoon:
`FiberRefs` propagation using `ThreadLocal`

### DIFF
--- a/docs/slf4j1-bridge.md
+++ b/docs/slf4j1-bridge.md
@@ -40,11 +40,6 @@ import zio.logging.consoleJsonLogger
 
 val logger = Runtime.removeDefaultLoggers >>> consoleJsonLogger() >+> Slf4jBridge.initialize
 ```
-
-To enable log annotations and span propagation to underlying code which using SLF4J api, 
-`Runtime.enableCurrentFiber` needs to be added to application bootstrap. 
-Use of this flag will negatively impact performance, so it is recommended to use it, only when you need this feature.
-
 <br/>
 
 **NOTE** You should either use `zio-logging-slf4j` to send all ZIO logs to an SLF4j provider (such as logback, log4j etc) OR `zio-logging-slf4j-bridge` to send all SLF4j logs to
@@ -76,7 +71,7 @@ object Slf4jBridgeExampleApp extends ZIOAppDefault {
   )
 
   override val bootstrap: ZLayer[ZIOAppArgs, Any, Any] =
-    Runtime.enableCurrentFiber ++ Runtime.removeDefaultLoggers >>> consoleJsonLogger(
+    Runtime.removeDefaultLoggers >>> consoleJsonLogger(
       ConsoleLoggerConfig(
         LogFormat.label(
           "name",
--- a/docs/slf4j1-bridge.md
+++ b/docs/slf4j1-bridge.md
@@ -41,6 +41,10 @@ import zio.logging.consoleJsonLogger
 val logger = Runtime.removeDefaultLoggers >>> consoleJsonLogger() >+> Slf4jBridge.initialize
 ```
 
+To enable log annotations and span propagation to underlying code which using SLF4J api, 
+`Runtime.enableCurrentFiber` needs to be added to application bootstrap. 
+Use of this flag will negatively impact performance, so it is recommended to use it, only when you need this feature.
+
 <br/>
 
 **NOTE** You should either use `zio-logging-slf4j` to send all ZIO logs to an SLF4j provider (such as logback, log4j etc) OR `zio-logging-slf4j-bridge` to send all SLF4j logs to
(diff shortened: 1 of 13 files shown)


docs/slf4j1-bridge.md
```diff
@@ -41,6 +41,10 @@ import zio.logging.consoleJsonLogger
 val logger = Runtime.removeDefaultLoggers >>> consoleJsonLogger() >+> Slf4jBridge.initialize
 ```
 
+To enable log annotations and span propagation to underlying code which using SLF4J api,
```
adamgfraser, line 44: Can we build the current fiber integration integration into an existing layer such as `Slf4jBridge.initialize`? It is great to have the ability to customize when necessary but it is also good to make common use cases easy and `enableCurrentFiber` is really an implementation detail where if the user is going to do something that requires it (basically propagating context from `FiberRef` values to `ThreadLocal` values we should just do it.

slf4j-bridge/src/main/scala/org/slf4j/impl/ZioLoggerFactory.scala
```diff
@@ -43,7 +43,13 @@ class ZioLoggerFactory extends ILoggerFactory {
   private[impl] def run(f: ZIO[Any, Nothing, Any]): Unit =
     if (runtime != null) {
       zio.Unsafe.unsafe { implicit u =>
-        runtime.unsafe.run(f)
+        runtime.unsafe.run {
+          val fiberRefs = Fiber.currentFiber().map(_.asInstanceOf[Fiber.Runtime[_, _]].unsafe.getFiberRefs())
```
adamgfraser, line 47: You are within the ZIO name space so you could just use `_currentFiber` here to avoid the need for the cast. I will also look at specializing the return type of `currentFiber`.

slf4j-bridge/src/main/scala/org/slf4j/impl/ZioLoggerFactory.scala
```diff
@@ -43,7 +43,13 @@ class ZioLoggerFactory extends ILoggerFactory {
...
+          fiberRefs match {
+            case Some(fiberRefs) => ZIO.setFiberRefs(fiberRefs) *> f
```
adamgfraser, line 49: This pattern of running a ZIO workflow just to log something is not very efficient. We can get the current loggers from the current `FiberRefs` and then we can just use those loggers to log directly without going through a ZIO workflow.

adamgfraser (changes requested):
Much cleaner without the need for any new operators!

Made a couple of suggestions for further improvements.

slf4j-bridge/src/main/scala/org/slf4j/impl/ZioLoggerFactory.scala
justcoon: yes, i think you are right, we can probably do it better, 
probably by using `Fiber.log`
but i propose to do it in next PR

adamgfraser: 👍

slf4j-bridge/src/main/scala/org/slf4j/impl/ZioLoggerFactory.scala
justcoon: i changed it slf4j v2 bridge to use `_currentFiber`, as it is in zio namespace

docs/slf4j1-bridge.md
justcoon: i think we can put it to `Slf4jBridge.initialize` 
however there should be option to disable that as `enableCurrentFiber` may have impact to performance

which probably ends in something like

```scala
object Slf4jBridge {

  def initialize(propagateFibers: Boolean = true): ZLayer[Any, Nothing, Unit] = ???
```

adamgfraser: Makes sense. I might split it into two layers so it can be a `val` and you don't need the empty parentheses when you call it.

`
object Slf4jBridge {

  val initialize: ZLayer[Any, Nothing, Unit] =
    ???

  val initializeWithoutFiberRefPropagation: ZLayer[Any, Nothing, Unit] =
    ???
}
```

justcoon: @adamgfraser 

i added layers as you proposed, probably it is good that this propagation is enabled by default,
however it is also changing default behavior

adamgfraser: @justcoon Sounds good.My thought is that `EnableCurrentFiber` is necessary for integration between `FiberRef` and `ThreadLocal` values and that is exactly what you are asking us to do with the `Slf4jBridge` so that makes sense as the default and will give the "correct", though potentially not as performant value, so that should be the default. But if you think we should do it the other way definitely open to your thoughts on that.

Otherwise looks good to me subject to my comment about avoiding unnecessary ZIO workflows which we will handle separately.

adamgfraser (approved):

docs/slf4j1-bridge.md
justcoon: thank you very much for hints and review

PR #719, opened 2023-05-17
slf4j bridge - more direct logging

justcoon:
using `zio.internal.FiberRuntime` instead of `ZIO.log`


adamgfraser (approved):
Looks great!

PR #811, opened 2024-02-17
slf4j - cause to throwable using FiberFailure to have proper stacktrace

justcoon:


slf4j/src/main/scala/zio/logging/slf4j/SLF4J.scala
```diff
@@ -37,6 +37,16 @@ import java.util
 
 object SLF4J {
 
+  /**
+   * convert [[zio.Cause]] to [[java.lang.Throwable]] using [[zio.FiberFailure]]
+   */
+  val causeToThrowableDefault: Cause[Any] => Option[Throwable] = cause =>
```
adamgfraser, line 43: I wonder if this should just be `FiberFailure(cause)` and return a `Throwable` instead of an `Option[Throwable]`. A `Cause.empty` still means some failure occurred, it just means we don't have additional information about that failure.

justcoon: in current `zio.ZLogger` interface, `Cause` is always provided

```scala
trait ZLogger[-Message, +Output] { self =>
  def apply(
    trace: Trace,
    fiberId: FiberId,
    logLevel: LogLevel,
    message: () => Message,
    cause: Cause[Any],
    context: FiberRefs,
    spans: List[LogSpan],
    annotations: Map[String, String]
  ): Output
}
```

for example `ZIO.logInfo("Start")`:

Screenshot 2024-02-18 at 11 01 24

considering `zio.Cause.isEmpty` implementation, is it better in case of `causeToThrowableDefault` to pattern match on `zio.Cause.Empty` ?

adamgfraser: Makes sense.

adamgfraser (approved):

PR #881, opened 2024-09-03
Update formatting-log-records.md

neo773:
Required for https://github.com/zio/zio/issues/9162 as this document breaks with MDX v2


justcoon (approved):

neo773:
Looks like we need to bump the version to update the npm package.

PR #893, opened 2024-10-09
Add appendLoggerName aspect

somdoron:
We are upgrading from ZIO 1, where the previous version of zio.logging appended the logger name to the existing logger name. In the new version, it replaces it. This PR suggests introducing a new aspect, appendLoggerName, to append a string to the existing logger name.

Our codebase relies on setting the logger name. For example, when our RPC framework receives an incoming message, we add the service name and method name to the logger name.

### DIFF
--- a/core/shared/src/main/scala/zio/logging/package.scala
+++ b/core/shared/src/main/scala/zio/logging/package.scala
@@ -54,6 +54,23 @@ package object logging extends LoggerLayers {
   def loggerName(value: String): ZIOAspect[Nothing, Any, Nothing, Any, Nothing, Any] =
     ZIOAspect.annotated(loggerNameAnnotationKey, value)
 
+  /**
+   * Append logger name aspect, by which it is possible to append a logger name to the current logger name.
+   * The new name is appended using a dot (.) as the delimiter.
+   *
+   * annotation key: [[zio.logging.loggerNameAnnotationKey]]
+   */
+  def appendLoggerName(value: String): ZIOAspect[Nothing, Any, Nothing, Any, Nothing, Any] =
+    new ZIOAspect[Nothing, Any, Nothing, Any, Nothing, Any] {
+      def apply[R, E, A](zio: ZIO[R, E, A])(implicit trace: Trace): ZIO[R, E, A] =
+        for {
+          annotations      <- FiberRef.currentLogAnnotations.get
+          currentLoggerName = annotations.get(loggerNameAnnotationKey)
+          newLoggerName     = currentLoggerName.fold(value)(currentLoggerName => s"$currentLoggerName.$value")
+          a                <- ZIO.logAnnotate(loggerNameAnnotationKey, newLoggerName)(zio)
+        } yield a
+    }
+
   implicit final class LogAnnotationZIOSyntax[R, E, A](private val self: ZIO[R, E, A]) {
     def logAnnotate[V: Tag](key: LogAnnotation[V], value: V): ZIO[R, E, A] =
       self @@ key(value)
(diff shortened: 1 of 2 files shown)


justcoon:
hello @somdoron , 
there is already logger name annotation https://github.com/zio/zio-logging/blob/master/core/shared/src/main/scala/zio/logging/package.scala#L54 
is not that what you are looking for ?

core/shared/src/main/scala/zio/logging/package.scala
```diff
@@ -54,6 +54,23 @@ package object logging extends LoggerLayers {
...
+   * Append logger name aspect, by which it is possible to append a logger name to the current logger name.
+   * The new name is appended using a dot (.) as the delimiter.
+   *
+   * annotation key: [[zio.logging.loggerNameAnnotationKey]]
+   */
+  def appendLoggerName(value: String): ZIOAspect[Nothing, Any, Nothing, Any, Nothing, Any] =
+    new ZIOAspect[Nothing, Any, Nothing, Any, Nothing, Any] {
+      def apply[R, E, A](zio: ZIO[R, E, A])(implicit trace: Trace): ZIO[R, E, A] =
+        for {
+          annotations      <- FiberRef.currentLogAnnotations.get
```
justcoon, line 67: i think there is `ZIO.logAnnotations` function , which actually calling `FiberRef.currentLogAnnotations.get` but it seems to be better to use `ZIO.logAnnotations` here

somdoron: thanks, fixed.

core/shared/src/main/scala/zio/logging/package.scala
```diff
@@ -54,6 +54,23 @@ package object logging extends LoggerLayers {
   def loggerName(value: String): ZIOAspect[Nothing, Any, Nothing, Any, Nothing, Any] =
     ZIOAspect.annotated(loggerNameAnnotationKey, value)
 
+  /**
+   * Append logger name aspect, by which it is possible to append a logger name to the current logger name.
+   * The new name is appended using a dot (.) as the delimiter.
+   *
+   * annotation key: [[zio.logging.loggerNameAnnotationKey]]
+   */
+  def appendLoggerName(value: String): ZIOAspect[Nothing, Any, Nothing, Any, Nothing, Any] =
```
justcoon, line 63: just thinking about aspect like

```scala
def setLoggerName(fn: Option[String] => String): ZIOAspect[Nothing, Any, Nothing, Any, Nothing, Any] = ...
```
(not sure if it is the best name ...)

and then
```scala
def appendLoggerName(value: String): ZIOAspect[Nothing, Any, Nothing, Any, Nothing, Any] =
   setLoggerName(currentLoggerName => currentLoggerName.fold(value)(currentLoggerName => s"$currentLoggerName.$value"))
```

somdoron: make sense, regarding the name, maybe just `loggerName`? It would have different overload than the current one.

justcoon: @somdoron what do you think about that ?

justcoon: not bad idea to use `loggerName`, even if overloads sometimes making problems

somdoron: I pushed a commit with your suggestion, let me know what you think

justcoon: nice,

justcoon:
@somdoron thanks for contribution, i have to take a look on CI issues first ..., as builds are failing currently

justcoon:
hi @somdoron, i think you need to run scalafix

justcoon (approved):

PR #896, opened 2024-10-17
Update README.md

felix-hedenstrom:
Change suggested version


justcoon:
hello @felix-hedenstrom , thanks for contribution, 
actually CI triggering pipeline which should make PR, if version changed
https://github.com/zio/zio-logging/pull/898
however it is not triggered with release, but with next code change
just note, there is version on multiple places ...

PR #909, opened 2024-11-23
build(core): support scala native + enabled JS tests

ThijsBroersen:
This PR provides support for Scala Native for module `core`. 

While implementing I discovered all tests were only defined for the JVM. For JS nothing was tested. I moved all tests to shared.
Then I discovered all File related code is incompatible with the JS platform. I moved this to jvm-native.

Only the mimaChecks seem fail because I had to move some shared zio.logging package code and add the platform specific package code in the platform directories. 
I could fix it by duplicating the package objects.. 
Is there any way it can be fixed without duplicating the code? I tried the scala3-library transparentTrait but that did not fix it and is not available for Native.


examples/core/shared/src/main/scala/zio/logging/example/SimpleApp.scala
```diff
@@ -15,18 +15,20 @@
...
+import zio.internal.Platform
 import zio.logging.{ ConsoleLoggerConfig, consoleLogger }
-import zio.{ ExitCode, Runtime, Scope, ZIO, ZIOAppArgs, ZIOAppDefault, ZLayer }
 
 object SimpleApp extends ZIOAppDefault {
 
   override val bootstrap: ZLayer[ZIOAppArgs, Any, Any] =
-    Runtime.removeDefaultLoggers >>> consoleLogger(ConsoleLoggerConfig.default)
+    if (Platform.isJS) ZLayer.empty
+    else Runtime.removeDefaultLoggers >>> consoleLogger(ConsoleLoggerConfig.default)
```
ThijsBroersen, line 26: on JS this seems to remove all logging... ideas?

core/shared/src/main/scala/zio/logging/internal/LogAppender.scala
```diff
@@ -152,7 +152,7 @@ private[logging] object LogAppender {
   }
 
   def json(textAppender: String => Any): LogAppender = new LogAppender { self =>
-    val AnsiColorRegex = "\\u001b\\[\\d+m".r
+    val AnsiColorRegex = """\033\[\d+m""".r
```
ThijsBroersen, line 155: Not compatible with Native. This ANSI is.

justcoon:
hi @ThijsBroersen , thank you very much for contribution

examples/core/shared/src/main/scala/zio/logging/example/SimpleApp.scala
justcoon: personally i do not have experience with scala JS, but if there are differences i think is better to do example specific for jvm and JS

build.sbt
```diff
@@ -16,6 +16,7 @@ inThisBuild(
   List(
     name               := "zio-logging",
     ciEnabledBranches  := Seq("master"),
+    // ciJvmOptions ++= Seq("-Xmx6g", "-Xss2m", "-XX:+UseG1GC"),
```
justcoon, line 19: is this some leftover ?

ThijsBroersen: is was rendered in the ci.yaml but not properly picked up in the job. If you check the jobs here then you will see it was picked up but later in the job it still seemed to have only the default 1GB of memory: https://github.com/zio/zio-logging/actions/runs/11993647760/job/33434815940
So for now I added a `.jvmopts` file to fix it.

examples/core/shared/src/main/scala/zio/logging/example/SimpleApp.scala
justcoon: anyway, with that condition, it look like that for platform JS it will not use `consoleLogger(ConsoleLoggerConfig.default)`

ThijsBroersen: fixed it, all platforms use the same setup again. And discovered it wasn't working before because the scala-java-time-tzdb dependency was missing. Added them to 'core' because it is always required when using zio-logging.

core/shared/src/test/scala/zio/logging/ReconfigurableLoggerSpec.scala
```diff
@@ -49,13 +49,13 @@ object ReconfigurableLoggerSpec extends ZIOSpecDefault {
             _         <- ZIO.logDebug("debug")
             elements1 <- queue.takeAll
             _         <- TestSystem.putProperty("logger/format", "%level %message")
-            _         <- ZIO.sleep(500.millis)
+            _         <- ZIO.sleep(1000.millis)
```
ThijsBroersen, line 52: it looks like on the JS platform it sometimes takes a bit longer to reconfigure

justcoon:
@ThijsBroersen nice, actually did you look into https://github.com/zio/zio-logging/tree/master/docs section if something do not need to be updated? seems to not, but just in case ...

ThijsBroersen:
@justcoon update one line which was only mentioning JVM and JS support.

justcoon (approved):

justcoon:
hello @ThijsBroersen , thanks for contribution, can we follow with merge of this PR, or do you want to add something?
can you rebase your PR to latest master changes? thanks

ThijsBroersen:
> hello @ThijsBroersen , thanks for contribution, can we follow with merge of this PR, or do you want to add something?
> can you rebase your PR to latest master changes? thanks

All is fine. Not near a computer for the next few hours, can rebase later today.

ThijsBroersen:
Rebased

PR #922, opened 2025-01-04
zio 2.1.14, test fixes

justcoon:
zio 2.1.14, test fixes (Trace.unapply changed, and there were issues with line 0)

### DIFF
--- a/core/shared/src/test/scala/zio/logging/LoggerNameExtractorSpec.scala
+++ b/core/shared/src/test/scala/zio/logging/LoggerNameExtractorSpec.scala
@@ -29,7 +29,7 @@ object LoggerNameExtractorSpec extends ZIOSpecDefault {
             } else trace
           }
         val result      = extractor(
-          Trace.apply(trace, "", 0),
+          Trace.apply(trace, "", 1),
           FiberRefs.empty,
           annotations
         )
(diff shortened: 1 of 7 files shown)


core/shared/src/test/scala/zio/logging/LoggerNameExtractorSpec.scala
```diff
@@ -29,7 +29,7 @@ object LoggerNameExtractorSpec extends ZIOSpecDefault {
             } else trace
           }
         val result      = extractor(
-          Trace.apply(trace, "", 0),
+          Trace.apply(trace, "", 1),
```
guizmaii, line 32: Can I ask why you need to change these `0` into `1`? 🤔

justcoon: hi @guizmaii , look like there is some issue with latest change

```scala
import zio.test._
import zio.Trace

object TraceSpec extends ZIOSpecDefault {

  val spec: Spec[Environment, Any] = suite("TraceSpec")(
    test("line") {
      val expected = 0

      val trace = Trace.apply("", "", expected)

      val line = trace match {
        case Trace(_, _, line) => line
        case _                 => -1
      }

      assertTrue(line == expected)

    }
  )
}
```

fail with
```
Assertion failed:
Expected :0
Actual   :-1
<Click to see difference>

  ✗ -1 was not equal to 0
  line == expected
  line = -1
```
however, I guessed that in reality there will not be `Trace` with line=0, so I just updated tests

guizmaii: Interesting. I guess the previous behaviour was kind of incorrect as the `Trace.unapply` should probably have returned `None` for an empty `String`, don't you think? 🤔

Issue #923, opened 2025-01-05
Perf: Consider replacing the use of `Trace.unapply` with `Trace.parseOrNull`


guizmaii:
In ZIO 2.1.14, we've made some performance optimisations to the `Trace` methods.
See:
- https://github.com/zio/zio/pull/9414
- https://github.com/zio/zio/pull/9416
- https://github.com/zio/zio/pull/9433

The `Trace.parseOrNull` (package private to `zio` package) returns `null` if it can't parse the trace or returns a `ParsedTrace` case class (instead of a `Tuple`), which avoids boxing

I'm not sure how much you use `Trace.unapply` but I just wanted to make you aware of these changes and potential optimisation.

Feel free to close this ticket if irrelevant 🙂

justcoon:
hi @guizmaii , thanks, i will take a look on that

PR #924, opened 2025-01-05
optimisation in relation to Trace parse

justcoon:
fixes: #923

### DIFF
--- a/core/shared/src/main/scala/zio/logging/LogFormat.scala
+++ b/core/shared/src/main/scala/zio/logging/LogFormat.scala
@@ -15,7 +15,6 @@
  */
 package zio.logging
 
-import zio.internal.stacktracer.Tracer
 import zio.logging.internal._
 import zio.parser.{ Syntax, _ }
 import zio.{ Cause, Chunk, Config, FiberId, FiberRefs, LogLevel, LogSpan, Trace, ZLogger }
@@ -842,8 +841,8 @@ object LogFormat {
 
   val enclosingClass: LogFormat =
     LogFormat.make { (builder, trace, _, _, _, _, _, _, _) =>
-      val parsed = Tracer.instance.parseOrNull(trace)
-      if (parsed != null) {
+      val parsed = Trace.parseOrNull(trace)
+      if (parsed ne null) {
         builder.appendText(parsed.file)
       } else {
         builder.appendText("not-available")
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/logging/LogFormat.scala
```diff
@@ -841,9 +842,11 @@ object LogFormat {
 
   val enclosingClass: LogFormat =
     LogFormat.make { (builder, trace, _, _, _, _, _, _, _) =>
-      trace match {
-        case Trace(_, file, _) => builder.appendText(file)
-        case _                 => builder.appendText("not-available")
+      val parsed = Tracer.instance.parseOrNull(trace)
```
guizmaii, line 845: You can also use `Trace.parseOrNull`
```suggestion
      val parsed = Trace.parseOrNull(trace)
```

That's the "public" function used in the ZIO codebase. We don't usually use `Tracer.instance.xxx` directly

core/shared/src/main/scala/zio/logging/LogFormat.scala
```diff
@@ -841,9 +842,11 @@ object LogFormat {
...
+      if (parsed != null) {
```
guizmaii, line 846:
```suggestion
      if (parsed ne null) {
```

core/shared/src/main/scala/zio/logging/LogFormat.scala
```diff
@@ -868,9 +871,9 @@ object LogFormat {
     }
 
   val traceLine: LogFormat = LogFormat.make { (builder, trace, _, _, _, _, _, _, _) =>
-    trace match {
-      case Trace(_, _, line) => builder.appendNumeric(line)
-      case _                 => ()
+    val parsed = Tracer.instance.parseOrNull(trace)
+    if (parsed != null) {
```
guizmaii, line 875:
```suggestion
    if (parsed ne null) {
```

core/shared/src/main/scala/zio/logging/LoggerNameExtractor.scala
```diff
@@ -78,17 +79,19 @@ object LoggerNameExtractor {
...
-      case Trace(location, _, _) =>
-        val last = location.lastIndexOf(".")
-        val name = if (last > 0) {
-          location.substring(0, last)
-        } else location
-        Some(name)
-      case _                     => None
+  val trace: LoggerNameExtractor = FnExtractor { (trace, _, _) =>
+    val parsed = Tracer.instance.parseOrNull(trace)
+    if (parsed != null) {
```
guizmaii, line 84:
```suggestion
    if (parsed ne null) {
```

guizmaii (approved):

